PR #1, opened 2020-01-06
Generalize pit_loss to arbitrary shape and to retrieve the min perm

thequilo:
This PR generalizes `pit_loss` to use an arbitrary shape and axis.

It now uses python `min` instead of `torch.min` to get both the min loss and permutation at the same time. I am not sure if this is good or not; it makes the code easier to read but slower (I think).

### DIFF
--- a/padertorch/ops/losses/loss.py
+++ b/padertorch/ops/losses/loss.py
@@ -114,9 +114,20 @@ def pit_loss(
         tensor(1.)
 
         >>> T, K, F = 4, 2, 5
-        >>> estimate, target = torch.ones(K, F, T), torch.zeros(K, F, T)
+        >>> estimate = torch.stack([torch.ones(F, T), torch.zeros(F, T)])
+        >>> target = estimate[(1, 0), :, :]
         >>> pit_loss(estimate, target, axis=0, return_permutation=True)
-        (tensor(1.), (0, 1))
+        (tensor(1.), (1, 0))
+
+        >>> K = 5
+        >>> estimate, target = torch.ones(K), torch.zeros(K)
+        >>> pit_loss(estimate, target, axis=0)
+        tensor(1.)
+
+        >>> A, B, K, C, F = 4, 5, 3, 100, 128
+        >>> estimate, target = torch.ones(A, B, K, C, F), torch.zeros(A, B, K, C, F)
+        >>> pit_loss(estimate, target, axis=-3)
+        tensor(1.)
     """
     axis = axis % estimate.ndimension()
     sources = estimate.size()[axis]
--- a/padertorch/ops/losses/loss.py
+++ b/padertorch/ops/losses/loss.py
@@ -1,3 +1,4 @@
+from functools import partial
 import torch
 import torch.nn.functional
 from torch.distributions import Normal, MultivariateNormal
@@ -79,13 +80,18 @@ def deep_clustering_loss(x, t):
 def pit_loss(
         estimate: torch.Tensor,
         target: torch.Tensor,
+        axis: int,
         loss_fn=torch.nn.functional.mse_loss,
-        axis: int = -2,
         return_permutation: bool = False
 ):
-    """Does not support batch dimension. Does not support PackedSequence.
+    """
+    Permutation invariant loss function. Calls `loss_fn` on every possible
+    permutation between `estimate`s and `target`s and returns the minimum
+    loss among them. The tensors are permuted along `axis`.
 
-    Parameters:
+    Does not support batch dimension. Does not support PackedSequence.
+
+    Args:
         estimate: Padded sequence. The speaker axis is specified with `axis`,
             so the default shape is (T, K, F)
         target: Padded sequence with the same shape as `estimate` (defaults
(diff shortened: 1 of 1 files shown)


thequilo:
And I just noticed: I should add a doctest for more and less axis than 3

padertorch/ops/losses/loss.py
```diff
@@ -105,12 +131,19 @@ def pit_loss(estimate, target, loss_fn=torch.nn.functional.mse_loss):
             f'{estimate.size()} != {target.size()}'
         )
     candidates = []
+    filler = (slice(None),) * axis
```
boeddeker: Did you mean `filler = (slice(None),) * estimate.size()`?

padertorch/ops/losses/loss.py
```diff
@@ -105,12 +131,19 @@ def pit_loss(estimate, target, loss_fn=torch.nn.functional.mse_loss):
             f'{estimate.size()} != {target.size()}'
         )
     candidates = []
+    filler = (slice(None),) * axis 
     for permutation in itertools.permutations(range(sources)):
-        candidates.append(loss_fn(
-            estimate[:, permutation, :],
+        candidates.append((loss_fn(
+            estimate[filler + (permutation, )],
```
boeddeker: This looks wrong.
Wouldn't is be `filler[axis] = permutation` and `estimate[filler]`?

padertorch/ops/losses/loss.py
thequilo: No, I meant to add `axis %= estimate.ndimension()`. But there are now to possibilites to solve this. My version: `estimate[filler + (permutation, )]`, your version would be `filler[axis] = permutation` and then index with `estimate[filler]`

padertorch/ops/losses/loss.py
thequilo: Have a look at my previous comment, both are correct if `axis` is computed correctly

padertorch/ops/losses/loss.py
```diff
@@ -105,12 +131,19 @@ def pit_loss(estimate, target, loss_fn=torch.nn.functional.mse_loss):
...
             target
-        ))
-    return torch.min(torch.stack(candidates))
+        ), permutation))
+
+    min_loss, min_perm = min(candidates, key=lambda x: x[0])
```
boeddeker: `torch.min` has the advantage that the execution is async on the GPU.
Python `min` introduces a sync point. It would be nice to have no sync point.

thequilo: It is not possible to have no sync point if we want to output the permutation (right?). We could use `torch.argmax` and then index into the list of permutations in case the permutation is requested. Then we would at least skip the sync point when the permutation is not requested

padertorch/ops/losses/loss.py
boeddeker: Ah, you use another trick to get the axis position. Your solution is also fine.

padertorch/ops/losses/loss.py
boeddeker: You are right. When the permutation is needed, we have to sync the GPU and the CPU.
Since in many cases we do not need the permutation, `torch.argmax` would be nicer, because then pytorch has the chance to do the operation without a sync.

LukasDrude:
I suggest that `axis` is a mandatory argument. Otherwise `-2` seems to be a bit arbitrary default.

thequilo:
This could potentially break code that uses the old `pit_loss` or `pit_mse_loss` functions. We could do `pit_mse_loss = partial(pit_loss, axis=-2)`.

boeddeker:
I think a breaking change with such an error is ok.
In the doctest should then be a note that `axis=-2` will yield the old behavior.

thequilo:
Should `axis` then be a mandatory argument or should it have a default value like `axis=0`?

boeddeker:
I prefer to get the old behavior or the execution should fail.
Since `axis=0` may work or yield a strange error msg, I find `axis` as mandatory argument better.

thequilo:
Same for `pit_mse_loss` or should this be held compatible with using `partial(pit_loss, axis=-2)`? `pit_mse_loss` is already marked with `# this function is kept at the moment for backwards compatibility` so I think we should keep compatibility here

thequilo:
And how about `mse_loss` as default for `loss_fn`? This only makes sense for some PIT models, but there are many other cases where one might use different loss functions. I feel like from the name `pit_loss` the default argument `mse_loss` does not make too much sense.

boeddeker:
I think MSE as default is fine. The first paper where PIT is present, they used MSE.
And MSE is a common loss function.

padertorch/ops/losses/loss.py
```diff
@@ -105,16 +152,24 @@ def pit_loss(estimate, target, loss_fn=torch.nn.functional.mse_loss):
...
+
+    if return_permutation:
+        return min_loss, permutations[int(idx)]
+    else:
+        return min_loss
 
 # this function is kept at the moment for backwards compatibility
 # ToDo: remove this function
-pit_mse_loss = pit_loss
+pit_mse_loss = partial(pit_loss, axis=-2)
```
jensheit: I would suggest removing lines 169 to 172 and keep mse as default loss for pit

thequilo: `pit_mse_loss` is used in the `PermutationInvariantTraining` models. If we remove lines 169 to 172 we have to touch those models as well

thequilo: And some other people might use it in their personal repositories

jensheit: In that case I am fine with keeping it for the moment and removing it in a separate pull request together with all uses in padertorch.

boeddeker:
LGTM. I will merge.
Deleting `pit_mse_loss` will be done in another PR.

PR #3, opened 2020-01-13
Add LRSchedulerHook

thequilo:
Add a hook that applies a leanring rate scheduler to a training. This only supports schedulers that don't use metrics.

### DIFF
--- a/padertorch/train/hooks.py
+++ b/padertorch/train/hooks.py
@@ -13,6 +13,7 @@ from enum import IntEnum
 from pathlib import Path
 import types
 
+from distutils.version import LooseVersion
 import numpy as np
 import progressbar
 import torch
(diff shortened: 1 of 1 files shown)


padertorch/train/hooks.py
```diff
@@ -816,3 +817,43 @@ def close(self, trainer):
...
+    # It is very likely that this check is exclusive to this hook
+    IS_PYTORCH_1_1 = LooseVersion(torch.__version__) >= '1.1.0'
+
+    def __init__(self, lr_scheduler, trigger=(1, 'epoch')):
+        super().__init__(trigger)
+        self.lr_scheduler = lr_scheduler
+
+    def pre_step(self, trainer: 'pt.Trainer'):
+        if self.trigger(iteration=trainer.iteration, epoch=trainer.epoch):
+            if trainer.epoch > 0 or not self.IS_PYTORCH_1_1:
```
jensheit: should the step be called if pytorch larger 1.1.0 or if it is not?

thequilo: It should be skipped in the first epoch if pytorch is larger than 1.1.0

jensheit: so it should be an and?

jensheit: ah okay, to prevent the warning. Then it is fine

boeddeker:
Can you move the code below the `BackOffValidationHook`? This hook also does some kind of learning rate scheduling.

thequilo:
Yes, I'll open another pull request for it.

Btw, the file `hooks.py` is becoming quite large, might be worth thinking about how to split it.

thequilo:
Or maybe I don't because I can just push to this repository.

padertorch/train/hooks.py
thequilo: It does not only prevent the warning. The learning rate schedule would be shifted by one epoch otherwise

padertorch/train/hooks.py
```diff
@@ -816,3 +817,43 @@ def close(self, trainer):
...
+        ... ))  # doctest: +SKIP
+
+    Note:
+
+        This hook can only be used with learning rate schedulers that
+        don't require metrics.
+
+    """
+    # It is very likely that this check is exclusive to this hook
+    IS_PYTORCH_1_1 = LooseVersion(torch.__version__) >= '1.1.0'
```
boeddeker: Can you change this to something like `PYTORCH_ge_1_1`? Do you have an issue, where they mentioned this shift?

thequilo: Yes, and yes it's https://github.com/pytorch/pytorch/pull/7889

boeddeker: Thanks, could you add that reference to the issue to the source code?

thequilo: Done!

boeddeker: Thanks, I found them directly in the master.

PR #4, opened 2020-02-06
write hook states to checkpoint

JanekEbb:


padertorch/train/hooks.py
```diff
@@ -397,40 +407,27 @@ def __init__(
...
-            )
-        return self._json_file
-
     @property
     def _best_ckpt_name(self):
         return f"ckpt_best_{self.metric}.pth"
 
-    def save_validation_state(self):
-        validation_state = {
+    def state_dict(self):
```
jensheit: shouldn't the state_dict be a property of the hook?

boeddeker: Usually yes, but here we want to match pytorch `Module.state_dict()`:
https://github.com/pytorch/pytorch/blob/8195961f20caa18863698d85410d0818bfd67463/torch/nn/modules/module.py#L671
```python
def state_dict(self, destination=None, prefix='', keep_vars=False):
```

padertorch/train/hooks.py
```diff
@@ -39,18 +38,18 @@ class Priority(IntEnum):
...
+    Validation 20
+    Checkpoint 11
     End 10
 
     End has to be the last one
     Summary before Validation, clears timer information
     Print and ProgressBar may access Summary
     """
     END = 10
+    CHECKPOINT = 11
```
boeddeker: Should we write somewhere a motivation, why CHECKPOINT is the pre last Hook?

padertorch/train/hooks.py
```diff
@@ -61,6 +60,23 @@ class Hook:
...
 
+    @property
+    def uid(self):
+        """
+        A unique ID of a hook.
+
+        The default `uid` disallows to register a state full hook more than one
+        in the trainer. Hooks that may registered more than once should
+        overwrite this property (e.g. `ModelAttributeAnnealingHook`)
+        """
```
boeddeker: What do you think about such a docstring?

padertorch/train/hooks.py
```diff
@@ -358,6 +374,9 @@ def pre_step(self, trainer: 'pt.Trainer'):
     def close(self, trainer):
         self._save_latest_checkpoint(trainer)
 
+    def set_last(self, iteration, epoch):
+        pass
```
boeddeker: Why do you disable `set_last` for `CheckpointHook`?

boeddeker: May be problematic with Backoff.

padertorch/train/hooks.py
```diff
@@ -445,22 +449,18 @@ def finalize_summary(self, trainer):
...
+                  f' {trainer.iteration} iterations')
+            raise StopTraining
 
     def run_validation(self, trainer: 'pt.Trainer'):
         ckpt_dir = trainer.checkpoint_dir
-        if self._json_file is None:
-            self._json_file = ckpt_dir / self._json_filename
-            if self._json_file.exists():
-                self.load_validation_state()
         ckpt_path: Path = trainer.default_checkpoint_path()
```
boeddeker: Can you add a comment that `ckpt_path` does not exist at this moment, because it will be written later with the state of this hook?

padertorch/train/hooks.py
JanekEbb: Yes. Done.

padertorch/train/hooks.py
JanekEbb: This was to make CheckpointHook rewrite the checkpoint after backoff with updated remaining_backoffs. But you are right, this was not very clean. Now the CheckpointHook checks whether it is a backoff or resume and sets last accordingly such that a checkpoint is only rewritten after backoff.

padertorch/train/hooks.py
JanekEbb: Done

padertorch/train/hooks.py
JanekEbb: I like. Thanks.

padertorch/train/hooks.py
```diff
@@ -608,14 +622,17 @@ def _back_off(self, trainer: 'pt.Trainer'):
...
         best_iter = int(best_ckpt[len('ckpt_'): -len('.pth')])
         for j in range(1, len(self.ckpt_ranking)):
-            ckpt = self.ckpt_ranking[-j]
+            ckpt = self.ckpt_ranking[-j][0]
             if int(ckpt[len('ckpt_'): -len('.pth')]) > best_iter:
                 ckpt_path = ckpt_dir / ckpt
-                assert ckpt_path.exists(), ckpt_path
-                ckpt_path.unlink()
-                self.ckpt_ranking.pop(-j)
+                if ckpt_path.exists():
```
boeddeker: Did you change the `assert` to an `if` because the latest checkpoint does not exist?

boeddeker:
The change in the ValidationHook is a breaking change.
The training of old models cannot be resumed.
Do we need a backup plan to support resume for old models?

jensheit:
> The change in the ValidationHook is a breaking change.
> The training of old models cannot be resumed.
> Do we need a backup plan to support resume for old models?

I do not think that is necessary, since you probably do not update during training. If the problem occurs they can just add the missing keys to their checkpoints

LukasDrude:
You can change the checkpoint loading mechanism to also support the previous standard and make that clear with a proper comment.

padertorch/train/hooks.py
JanekEbb: yes

JanekEbb: added a comment

JanekEbb:
> You can change the checkpoint loading mechanism to also support the previous standard and make that clear with a proper comment.

Now it is possible to load old checkpoints with a warning that hook states are missing. If a hook state, e.g., the validation state, should be recovered, the checkpoint can still be easily rewritten manually as @jensheit suggested.

boeddeker:
I merged now this PR.

PR #5, opened 2020-02-13
collate_fn: support dataclasses

boeddeker:


boeddeker:
@jensheit Does the doctest covers all your use cases?

jensheit:
> @jensheit Does the doctest cover all your use cases?

yes it does

PR #9, opened 2020-03-20
Breaking change in Janek's feature augmentation

boeddeker:
When the `n` argument of `linear_warping` is an empty list, the function no longer assumes that the value was `1` (i.e. adding a singleton dimension).


boeddeker:
Now all tests pass.
The failures in the tests came from changed dependencies in paderbox (https://github.com/fgnt/paderbox/pull/11) and a change of `paderbox.notebook.pprint` (https://github.com/fgnt/paderbox/pull/9).

I will merge this PR.

PR #12, opened 2020-05-15
Move DPRNN to padertorch

thequilo:
This PR moves the DPRNN implementation from padercontrib to padertorch. It still needs a little bit of cleanup and testing.

Things that are not that clear yet:
 - where to put the chunking: contrib/neumann or data?
 - put the training script in contrib or examples?


jensheit:
In my opinion at least the train and evaluate script should be moved to contrib.examples.dual_path_rnn. The same is true for the README.md. One could argue that the chunking could stay in the current path.

jensheit:
In my opinion at least the train and evaluate script should be moved to contrib.examples.dual_path_rnn. The same is true for the README.md. One could argue that the chunking could stay in the current path.

jensheit:
We should think of a better name for the model and modules, perhaps we could create an additional directory modules/tasnet/encoder modules/tasnet/decoder and change the model to dprnn_source_separation or create additonal sub directories here as well.

padertorch/models/dual_path_rnn_source_separator.py
```diff
@@ -0,0 +1,278 @@
...
+
+    return {
+        'relu': torch.nn.ReLU(),
+        'sigmoid': torch.nn.Sigmoid(),
+        'softmax': torch.nn.Softmax(dim=speaker_dim),
+        None: identity,
+        'None': identity,
+        'identity': identity,
+    }[nonlinearity]
+
```
jensheit, line 26: For get_nonlinearities you should use the pt.ops.mapping functions

padertorch/models/dual_path_rnn_source_separator.py
```diff
@@ -0,0 +1,278 @@
...
+        return {k: torch.mean(torch.stack(v)) for k, v in losses.items()}
+
+    def review(self, inputs, outputs):
+        # Report audios
+        audios = {
+            'observation': pt.summary.audio(
+                signal=inputs['y'][0], sampling_rate=8000
+            ),
+        }
+
```
jensheit, line 261: pt.summary.audio does not exist, probably you did not commit your change to the __init__.py

jensheit, line 261: pt.summary.audio does not exist, probably forgot to commit a change to all in tbx_utils

padertorch/ops/losses/__init__.py
```diff
@@ -1,3 +1,4 @@
 from . import loss
+from . import time_domain_losses
```
jensheit, line 2: if we add a time_domain_losses.py we probably have to rename the old loss or/and split it into multiple files

padertorch/models/dual_path_rnn_source_separator.py
thequilo: `pt.summary.audio` is a quite old change. It exists in `padertorch.summary.tbx_utils` and is imported to `pt.summary`. If not, something is very wrong with the git repository.

jensheit: I have pulled your request and I had to add audio to the __all__[] command at the begining of the tbx_utils.py file to get the script to work

padertorch/ops/losses/__init__.py
thequilo: I'm not sure what the best naming is here. I first put these losses into `loss`, but the file becomes pretty large then.

padertorch/models/dual_path_rnn_source_separator.py
thequilo: Hmm, works for me without that. I'll add it to `__all__`

padertorch/contrib/examples/tasnet/README.md
```diff
@@ -0,0 +1,74 @@
...
+| `with log_mse`  | 16.7  | 17.0  |
+
+Training
+--------
+
+A storage root must be set with `export STORAGE=/path/to/your/storage`.
+After installing `padercontrib`, a training can be started with
+
+```bash
+$ python -m padercontrib.pytorch.examples.dual_path_rnn.train
```
boeddeker, line 24: Can you change this to `padertorch` instead of `padercontrib`

thequilo:
The DPRNN-TasNet example should work fine now except for the dependency on `padercontrib.database.wsj0_mix.WSJ0_2Mix_8k`. What should we do about that?

boeddeker:
Can you reduce the dependency, so that you only depend on the json that is used in `padercontrib.database.wsj0_mix.WSJ0_2Mix_8k` and not the class?
Most of the properties are already public in the super class: `lazy_dataset_database.JsonDatabase`

thequilo:
I removed the import from padercontrib. The script now directly uses `lazy_dataset.database.JsonDatabase` and the path to the JSON has to be set in the configuration.

padertorch/contrib/examples/tasnet/train.py
```diff
@@ -0,0 +1,309 @@
...
+    iterator = db.get_dataset(dataset)
+
+    if iterator_slice is not None:
+        iterator = iterator[iterator_slice]
+
+    iterator = (
+        iterator
+        .map(pre_batch_transform)
+        .map(RandomChunkSingle(chunk_size, chunk_keys=('y', 's'), axis=-1))
+        .shuffle(reshuffle=False)
```
boeddeker, line 157: Is there a reason, why you disabled reshuffle? Here it should be supported.

padertorch/contrib/examples/tasnet/train.py
```diff
@@ -0,0 +1,309 @@
...
+        prepare_iterable_captured(db, validate_dataset),
+    )
+
+    # Register hooks and start the actual training
+    trainer.register_validation_hook(
+        prepare_iterable_captured(db, validate_dataset)
+    )
+
+    # Learning rate scheduler
+    trainer.register_hook(pt.train.hooks.LRSchedulerHook(
```
boeddeker, line 272: I think @jensheit removed the `pt.train.hooks.LRSchedulerHook` in favor of the BackOffValidationHook

jensheit: Yes, and the BackOffValidationHook should be added with register_validation_hook()

thequilo: But those two do different things. This one is as close to the original paper as possible. I don't know how they compare in terms of training behavior, but the backoff validation hook might be preferable

padertorch/contrib/examples/tasnet/train.py
thequilo: No, there's no reason for that. I think this was copied from the pit example

boeddeker: Can you change this to `reshuffle=True`? The PIT example was written, before https://github.com/fgnt/lazy_dataset/pull/2 was merged.

thequilo: Done!

padertorch/contrib/examples/tasnet/train.py
boeddeker: The `pt.train.hooks.LRSchedulerHook` is still available.
In #11 only LRSchedulers in `pt.train.optimizer` were deleted, `pt.train.hooks.LRSchedulerHook` is an alternative to the LRSchedulers in `pt.train.optimizer` and is now the recommended one, when no backoff is used/needed.

PR #14, opened 2020-05-19
Revised PIT example in contrib/examples

TCord:
Mainly removed dependencies of internal libraries in the PIT example.


padertorch/contrib/examples/pit/__init__.py
```diff
@@ -0,0 +1,49 @@
+from .model import PermutationInvariantTrainingModel
+
+MAKEFILE_TEMPLATE_TRAIN = """
```
boeddeker, line 3: Can you remove the code from the `__init__.py`?
It is always confusing, when you search the code.
Most of the time you only find the imports in the `__init__.py`

padertorch/contrib/examples/pit/data.py
```diff
@@ -0,0 +1,83 @@
...
+        db, dataset: str, batch_size, return_keys=None, prefetch=True,
+):
+    audio_keys = ['observation', 'speech_source']
+    iterator = db.get_dataset(dataset)
+
+    iterator = (
+        iterator
+        .map(partial(read_audio, audio_keys=audio_keys))
+        .map(partial(pre_batch_transform, return_keys=return_keys))
+        .shuffle()
```
boeddeker, line 21: Can you change `reshuffle=False` to `reshuffle=True`?

padertorch/contrib/examples/pit/data.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+def pre_batch_transform(inputs, return_keys=None):
+    s = inputs['audio_data']['speech_source']
+    y = inputs['audio_data']['observation']
+    S = stft(s, 512, 128)
+    Y = stft(y, 512, 128)
+    Y = einops.rearrange(Y, 't f -> t f')
+    S = einops.rearrange(S, 'k t f -> t k f')
+    X = S  # Same for MERL database
```
boeddeker, line 56: Can you remove the name MERL? This was our internal name. In public, it is mostly called wsj0_mix.

padertorch/train/hooks.py
```diff
@@ -168,6 +168,8 @@ def empty_summary_dict():
             texts=dict(),
             figures=dict(),
             timings=dict(),
+            buffers=dict(),
+            snapshots=dict()
```
boeddeker, line 172: Can you add a few comments, what buffers and snapshots are?

padertorch/contrib/examples/pit/__init__.py
TCord: I moved the templates to an additional file.

padertorch/train/hooks.py
TCord: I wrote a line on the two keys in the SummaryHook description. I think further documentary of those two keys should be placed inside the ModifySummary function.

padertorch/contrib/examples/pit/evaluate.py
```diff
@@ -92,6 +54,10 @@
 @ex.config
 def config():
     debug = False
+    db_path = ''
+    if "WSJ0_2MIX" in os.environ:
+        db_path = os.environ.get("WSJ0_2MIX")
+    assert len(db_path) > 0, 'Set path to database Json on the command line or set environment variable WSJ0_2MIX'
```
thequilo, line 60: You can raise a `sacred.utils.MissingConfigError` here to make clear that this config value is not set. Not sure which one is clearer to people that don't know sacred.

padertorch/contrib/examples/pit/evaluate.py
```diff
@@ -154,15 +119,15 @@ def init(_config, _run):
 
 
 @ex.main
-def main(_run, batch_size, datasets, debug, experiment_dir):
+def main(_run, batch_size, datasets, debug, experiment_dir, db_path):
```
thequilo, line 122: We might want to use the same name for `db_path` in this example and the DPRNN example. In the DPRNN it is called `database_json`, and I think this better describes the value to put here than `db_path`.

padertorch/contrib/examples/pit/train.py
```diff
@@ -133,12 +118,14 @@ def init(_config, _run):
 
 
 @ex.capture
-def prepare_and_train(_config, _run, train_dataset, validate_dataset):
+def prepare_and_train(_config, _run, train_dataset, validate_dataset, db_path):
```
thequilo, line 121: Same comment as in evaluate: We might want to ave the same name for `db_path` in here and the DPRNN example

padertorch/contrib/examples/pit/evaluate.py
TCord: I think it's nicer to do it as it is, for now, as you can understand the code in this way without knowing sacred

padertorch/contrib/examples/pit/evaluate.py
```diff
@@ -154,15 +119,15 @@ def init(_config, _run):
...
     if IS_MASTER:
         sacred.commands.print_config(_run)
 
     # TODO: Substantially faster to load the model once and distribute via MPI
     model = get_model()
-    db = MerlMixtures()
+    db = JsonDatabase(json_path=database_json)
 
     model.eval()
     with torch.no_grad():
```
jensheit, line 133: This should not be necessary

padertorch/contrib/examples/pit/evaluate.py
```diff
@@ -193,7 +160,7 @@ def main(_run, batch_size, datasets, debug, experiment_dir):
                 s = s[:, :z.shape[1]]
                 z = z[:, :s.shape[1]]
                 entry['mir_eval'] \
-                    = pb.evaluation.mir_eval_sources(s, z, return_dict=True)
+                    = pb_bss.evaluation.mir_eval_sources(s, z, return_dict=True)
```
jensheit, line 163: You should probably use all evaluations possible in pb_bss.evaluation since that is one of the benefits of pb_bss

padertorch/contrib/examples/pit/model.py
```diff
@@ -0,0 +1,152 @@
...
+        self.blstm = torch.nn.LSTM(
+            F,
+            units,
+            recurrent_layers,
+            bidirectional=True,
+            dropout=dropout_hidden,
+        )
+
+        assert dropout_linear <= 0.5, dropout_linear
+        self.dropout_linear = torch.nn.Dropout(dropout_linear)
```
jensheit, line 68: This could also be done by pt.module.fully_connected_stack

padertorch/contrib/examples/pit/train.py
```diff
@@ -20,36 +21,17 @@
...
-\t\t--res=rset=1:ncpus=4:gtx1080=1:ompthreads=1 \\
-\t\t--time=100h \\
-\t\t--join \\
-\t\t--stdout=stdout \\
-\t\t--tracefile=%x.%reqid.trace \\
-\t\t-N train_{nickname} \\
-\t\tpython -m {main_python_path} with config.json
-"""
-
+from padertorch.contrib.examples.pit.templates import MAKEFILE_TEMPLATE_TRAIN as MAKEFILE_TEMPLATE
```
jensheit, line 34: If possible you should try not to use anything from one of the personal contrib directories here

padertorch/train/hooks.py
boeddeker: Yes that would be a good place. Can you add there some sentence to describe these keys?
An assert in that function would also be nice, because then the user knows where to change it.

Is it possible to write a small test for buffers and snapshots? A "burn" test would be enough.

TCord: I will commit those in an additional issue, as the updates for hooks where not meant to be in this pull request, anyway.

padertorch/contrib/examples/pit/train.py
TCord: Removing the personal contrib function would make rewriting the example init neccessary. In my opinion, rewriting the example to remove the function would be too time consuming, as all the function does is the automatic creation of an empty folder.

padertorch/contrib/examples/pit/model.py
TCord: As it is intended as a simple example, I think the usage of several linear layers instead of the fully connected stack should be sufficient, here

padertorch/train/hooks.py
```diff
@@ -224,6 +232,12 @@ def _to_list(scalars):
             scalars = [scalars]
         return scalars
 
+    @staticmethod
+    def _detach(buffer):
+        if torch.is_tensor(buffer):
+            buffer.detach()
```
boeddeker, line 238: I think detach is no inplace operation.

PR #16, opened 2020-05-20
Add copy flag to to_numpy and make resulting array read-only by default

thequilo:
On CPU, the result of `pt.utils.to_numpy` sometimes shared the memory with the input tensor which caused issues in `pt.summary.audio` changing the input tensor.

### DIFF
--- a/padertorch/summary/tbx_utils.py
+++ b/padertorch/summary/tbx_utils.py
@@ -139,8 +139,8 @@ def audio(signal, sampling_rate: int = 16000, batch_first=False,
 
     # Normalize so that there is no clipping
     if normalize:
-        signal /= np.max(np.abs(signal))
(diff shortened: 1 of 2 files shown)


padertorch/summary/tbx_utils.py
```diff
@@ -139,8 +139,8 @@ def audio(signal, sampling_rate: int = 16000, batch_first=False,
 
     # Normalize so that there is no clipping
     if normalize:
-        signal /= np.max(np.abs(signal))
-        signal *= 0.95
+        signal = signal / np.max(np.abs(signal))
+        signal = signal * 0.95
```
boeddeker, line 143: Can you change this back to an inplace operation? Only the first shouldn't be inplace. For the second operation you can use inplace

PR #17, opened 2020-05-20
Warn if CUDA is not available in Trainer.train

thequilo:


padertorch/train/trainer.py
```diff
@@ -212,6 +212,26 @@ def train(
                 If None, the device of the model will not be changed and the
                 example to the model is moved to the cpu.
         """
+
+        if torch.cuda.is_available():
```
boeddeker, line 216: Can you change this to:
```python
if torch.cuda.is_available():
    if device is None:
        device = 0
```

PR #19, opened 2020-05-26
piece-wise linear annealing hooks (loss weight, model attribute, lear…

JanekEbb:
…ning rate)

### DIFF
--- a/padertorch/train/hooks.py
+++ b/padertorch/train/hooks.py
@@ -821,24 +821,31 @@ class AnnealingHook(TriggeredHook):
         """
         Base class for piece-wise linear annealing. The piece-wise linear
         function is parameterized by its breakpoints. It can also be used for
-        arbitrary annealing functions by stating breakpoints with an interval
+        arbitrary annealing functions when stating breakpoints with an interval
         similar to the trigger interval.
-        Before the first breakpoint there is a linear ramp between the initial
-        value and the first breakpoint. After the last breakpoint the function
-        stays constant at the value of the last breakpoint.
+        The annealing function is interpreted relative to the initial value,
+        i.e., a breakpoint (i, 1) corresponds to an absolute value equal to the
+        initial value of the parameter to be annealed.
+        Before the first breakpoint there is a linear connection between
+        (0, 1), which corresponds to the initial value, and the first breakpoint.
+        After the last breakpoint the function stays constant at the value of
+        the last breakpoint.
+        Note that you can still start with values differing from the initial
+        value by adding a breakpoint (0, y0).
 
         Args:
             trigger:
             breakpoints: list of (iteration, value) coordinates of the
-                piecewise linear funtion. Note that absolute (not relative)
-                values are expected. This, e.g., allows a linear ramp up from
-                0 to some value.
+                piecewise linear function. Note that breakpoint values are
+                interpreted relative to the initial value of the parameter
+                to be annealed.
             name: name of the attribute. You can use "attr1.attr11" to
                 anneal a sub attribute
         """
         super().__init__(trigger)
         self.breakpoints = sorted(breakpoints, key=lambda x: x[0])
         self.name = name
+        self.scale = None
         self.last_break = None
 
     @property
@@ -854,20 +861,22 @@ class AnnealingHook(TriggeredHook):
     def pre_step(self, trainer):
         if self.trigger(iteration=trainer.iteration, epoch=trainer.epoch):
             if self.last_break is None:
-                self.last_break = (0, self.get_value(trainer))
+                self.scale = self.get_value(trainer)
+                self.last_break = (0, 1)
             while len(self.breakpoints) > 0 and self.breakpoints[0][0] <= trainer.iteration:
                 self.last_break = self.breakpoints.pop(0)
             if len(self.breakpoints) > 0:
                 slope = (
-                        (self.breakpoints[0][1]-self.last_break[1])
-                        / (self.breakpoints[0][0]-self.last_break[0])
+                    (self.breakpoints[0][1]-self.last_break[1])
+                    / (self.breakpoints[0][0]-self.last_break[0])
                 )  # a = (y1 - y0) / (x1 - x0)
                 value = (
-                        self.last_break[1]
-                        + slope * (trainer.iteration - self.last_break[0])
+                    self.last_break[1]
+                    + slope * (trainer.iteration - self.last_break[0])
                 )  # y = y0 + a * (x - x0)
+                value *= self.scale  # relative to absolute
             else:
-                value = self.last_break[1]
+                value = self.last_break[1] * self.scale
             self.set_value(trainer, value)
 
 
--- a/padertorch/train/hooks.py
+++ b/padertorch/train/hooks.py
@@ -948,11 +948,10 @@ class LRAnnealingHook(AnnealingHook):
         if self.name is not None:
             assert isinstance(optimizer, dict), type(optimizer)
             optimizer = optimizer[self.name]
-        else:
-            assert (
-                hasattr(optimizer, 'optimizer')
-                and hasattr(optimizer.optimizer, 'param_groups')
-            ), type(optimizer)
+        assert (
+            hasattr(optimizer, 'optimizer')
+            and hasattr(optimizer.optimizer, 'param_groups')
+        ), type(optimizer)
         return optimizer.optimizer
 
     def get_value(self, trainer):
--- a/padertorch/train/hooks.py
+++ b/padertorch/train/hooks.py
@@ -32,6 +32,9 @@ __all__ = [
     'ProgressBarHook',
     'StopTrainingHook',
     'StopTraining',
+    'LossWeightAnnealingHook',
+    'ModelAttributeAnnealingHook',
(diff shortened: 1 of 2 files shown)


padertorch/train/hooks.py
```diff
@@ -812,105 +816,115 @@ class StopTraining(Exception):
...
-    Anneals a loss weight within the los_weights dict of the trainer.
-    """
-    def __init__(self, name, factor, trigger, max_value=None, min_value=None):
+class AnnealingHook(TriggeredHook):
+    def __init__(self, trigger, breakpoints, name):
         """
+        Base class for piece-wise linear annealing. The piece-wise linear
+        function is parameterized by its breakpoints. Can also be used for
+        arbitrary annealing functions by stating breakpoints with an interval
+        similar to the trigger interval.
```
boeddeker, line 825: Can you add a comment, what happens before the first breakpoint and what after the last breakpoint?

padertorch/train/hooks.py
```diff
@@ -812,105 +816,115 @@ class StopTraining(Exception):
...
         self.name = name
-        self.factor = factor
-        self.max_value = max_value
-        self.min_value = min_value
+        self.last_break = None
 
     @property
     def uid(self):
-        return super().uid + f"({self.name})"
+        return super().uid + f"({'.'.join(self.name)})"
```
boeddeker, line 840: Why do you use join here? Name should be already a string.

padertorch/train/hooks.py
```diff
@@ -812,105 +816,115 @@ class StopTraining(Exception):
...
+        if self.trigger(iteration=trainer.iteration, epoch=trainer.epoch):
+            if self.last_break is None:
+                self.last_break = (0, self.get_value(trainer))
+            while len(self.breakpoints) > 0 and self.breakpoints[0][0] <= trainer.iteration:
+                self.last_break = self.breakpoints.pop(0)
+            if len(self.breakpoints) > 0:
+                slope = (self.breakpoints[0][1]-self.last_break[1])/(self.breakpoints[0][0]-self.last_break[0])
+                value = self.last_break[1] + slope * (trainer.iteration - self.last_break[0])
+            else:
+                value = self.last_break[1]
```
boeddeker, line 858: This code is relative complicated.
Is my understanding correct: Take the original value as start values and then interpolate to the next break point?
i.e. the breakpoints are absolute and not relative the original value.

padertorch/train/hooks.py
```diff
@@ -812,105 +816,115 @@ class StopTraining(Exception):
...
+            trigger:
+            breakpoints:
+            name: states the key of the target optimizer when optimizer is a dict
+        """
+        super().__init__(trigger, breakpoints, name)
+
+    @property
+    def uid(self):
+        if self.name is None:
+            return super(AnnealingHook, self).uid
```
boeddeker, line 913: Can you add a comment, that you didn't write python 2 code here?

padertorch/train/hooks.py
JanekEbb: done

padertorch/train/hooks.py
JanekEbb: fixed

padertorch/train/hooks.py
JanekEbb: right, I added comments to improve clarity of the math

padertorch/train/hooks.py
JanekEbb: done

boeddeker:
Can you explain, why you used absolute values for the break points?
I expected relative values.

JanekEbb:
> Can you explain, why you used absolute values for the break points?
> I expected relative values.

Hm, not really. Maybe because its more verbose which values are used. Can you explain why you expected relative values?

boeddeker:
> Hm, not really. Maybe because its more verbose which values are used. Can you explain why you expected relative values?

When you read about learning rate decay, the people often write something like: after X epochs/iterations I reduce the learning rate by a factor of Y.

JanekEbb:
> > Hm, not really. Maybe because its more verbose which values are used. Can you explain why you expected relative values?
> 
> When you read about learning rate decay, the people often write something like: after X epochs/iterations I reduce the learning rate by a factor of Y.

so you expected each breakpoint to be relative to the previous breakpoint? Anyway one reason to not use relative breakpoints is to allow ramp ups from 0 to some value. I added a comment.

boeddeker:
> so you expected each breakpoint to be relative to the previous breakpoint? Anyway one reason to not use relative breakpoints is to allow ramp ups from 0 to some value. I added a comment.

I expected more, relative to the initial value. When I change the learning rate, I must now also change the breakpoints. With relative to the initial value are also ramps from zero possible.

JanekEbb:
> > so you expected each breakpoint to be relative to the previous breakpoint? Anyway one reason to not use relative breakpoints is to allow ramp ups from 0 to some value. I added a comment.
> 
> I expected more, relative to the initial value. When I change the learning rate, I must now also change the breakpoints. With relative to the initial value are also ramps from zero possible.

When the inital value is 0 !?

boeddeker:
> When the inital value is 0 !?

When the default value is 0, that value is deactivated. But why not interpret this value as "scale"?
With
Breakpoint: (0, 0), (10, 1), (20, 0)
and an initial value ~0 you also get a ramp starting from zero.

boeddeker:
What is the opinion of the others (@jensheit @thequilo @tglarner @TCord)?
Is a relative or absolute value more practical/intuitive?

thequilo:
For me, absolute values are more intuitive. I don't know how this is used in the literature so this might not be true for everyone. I also agree that relative values can be more practical in some cases. How about supporting both?

JanekEbb:
> > When the inital value is 0 !?
> 
> When the default value is 0, that value is deactivated. But why not interpret this value as "scale"?
> With
> Breakpoint: (0, 0), (10, 1), (20, 0)
> and an initial value ~0 you also get a ramp starting from zero.

Now I see your point. Agree.

JanekEbb:
> For me, absolute values are more intuitive. I don't know how this is used in the literature so this might not be true for everyone. I also agree that relative values can be more practical in some cases. How about supporting both?

Sorry saw your comment only after the latest push. Do you think it is enough that the annealing function corresponds to absolute values if the initial value is set to 1 ?

thequilo:
Yes, I guess that's fine

padertorch/train/hooks.py
```diff
@@ -812,105 +816,136 @@ class StopTraining(Exception):
...
+
+
+class LRAnnealingHook(AnnealingHook):
+    """
+    Anneals an optimizer learning rate.
+    """
+    def __init__(self, trigger, breakpoints, name=None):
+        """
 
+        Args:
```
boeddeker, line 924: Can you either copy the docstring from AnnealingHook or mention that the signature description is in AnnealingHook?

boeddeker:
One last question: The breakpoint unit is iteration. Should the unit epoch also be supported?

thequilo:
I'd say yes, because with iterations we have to change the breakpoints when the batch-size changes. Also, I think that many papers use epoch to describe such an annealing.

JanekEbb:
Do you suggest to adopt the unit from trigger or to add a separate argument?

boeddeker:
> Do you suggest to adopt the unit from trigger or to add a separate argument?

Yes and yes.

I vote for a separate `unit` argument. i.e.
```python
breakpoints=[(1, 3), (2, 3)], unit='epoch'
```
and not
```python
breakpoints=[((1, 'epoch'), 3), ((2, 'epoch'), 3)]
```
(Do you mean something like that with `adopt the unit from trigger`?)

I think we should not support different units for the breakpoints.
That would make the implementation complicated and I do not see a usecase.

padertorch/train/hooks.py
```diff
@@ -812,105 +816,144 @@ class StopTraining(Exception):
...
+            assert isinstance(optimizer, dict)
+            optimizer = optimizer[self.name]
+        return optimizer.optimizer
+
+    def get_value(self, trainer):
+        return self.get_optimizer(trainer).param_groups[0]['lr']
+
+    def set_value(self, trainer, value):
+        optimizer = self.get_optimizer(trainer)
+        for param_group in optimizer.param_groups:
```
thequilo, line 958: Does this work if the optimizer is a `dict` but `self.name=None`?

thequilo: If not, you should assert this in `get_optimizer`

padertorch/train/hooks.py
```diff
@@ -946,8 +946,13 @@ def uid(self):
...
         optimizer = trainer.optimizer
         if self.name is not None:
-            assert isinstance(optimizer, dict)
+            assert isinstance(optimizer, dict), type(optimizer)
             optimizer = optimizer[self.name]
+        else:
+            assert (
+                hasattr(optimizer, 'optimizer')
+                and hasattr(optimizer.optimizer, 'param_groups')
+            ), type(optimizer)
```
thequilo, line 955: This assert can as well be performed when `self.name is not None:`, the same assertion should hold for the contents of the optimizer dict. Can you remove the `else:`?

PR #23, opened 2020-06-13
Add mnist_example_vae (work in progress)

FrederikRautenberg:


padertorch/contrib/examples/vae_mnist/mnist_vae_example.py
```diff
@@ -0,0 +1,195 @@
...
+    trainset = torchvision.datasets.MNIST(root='./data', train=True,
+                                          download=True, transform=tf)
+    trainloader = torch.utils.data.DataLoader(trainset, batch_size=batch_size,
+                                              shuffle=True, num_workers=2)
+
+    testset = torchvision.datasets.MNIST(root='./data', train=False,
+                                         download=True, transform=tf)
+    testloader = torch.utils.data.DataLoader(testset, batch_size=batch_size,
+                                             shuffle=False, num_workers=2)
+
```
jensheit, line 50: I think we should at least entertain the idea of using lazy_dataset here instead of the DataLoader because this allows for much more freedom in preprocessing and it does not lead to discarding examples if len(dataset)%batch_size !=0

In padertorch.testing.test_db there is already a database object for MNIST which could be used here.

padertorch/contrib/examples/vae_mnist/mnist_vae_example.py
```diff
@@ -0,0 +1,195 @@
...
+    return trainloader, testloader
+
+
+class PadertorchModel(padertorch.base.Model):
+    MSE_loss = nn.MSELoss()
```
jensheit, line 55: this should be lower case following PEP 8 rules

padertorch/contrib/examples/vae_mnist/mnist_vae_example.py
```diff
@@ -0,0 +1,195 @@
...
+
+    def __init__(self, net):
+        super().__init__()
+        self.net = net
```
jensheit, line 59: why are you not directly integrating the vae class here? This would only make sense If you intend to replace the VAE with an other module. In this instance it just makes the code harder to read.

padertorch/contrib/examples/vae_mnist/mnist_vae_example.py
```diff
@@ -0,0 +1,195 @@
...
+
+    testset = torchvision.datasets.MNIST(root='./data', train=False,
+                                         download=True, transform=tf)
+    testloader = torch.utils.data.DataLoader(testset, batch_size=batch_size,
+                                             shuffle=False, num_workers=2)
+
+    return trainloader, testloader
+
+
+class PadertorchModel(padertorch.base.Model):
```
jensheit, line 54: This name is not informative. Either call it VaeModel if you directly integrate the VAE or you call it after the task it is supposed to perform.
Additionally use padertorch.Model and not padertorch.base.Model

padertorch/contrib/examples/vae_mnist/mnist_vae_example.py
```diff
@@ -0,0 +1,195 @@
...
+using the mnist database, sacred and the padertorch trainer
+
+Use the following command:
+    python mnist_vae_example.py with logfilepath=/path/to/dir/ epochs=500
+
+Other options for sacred are available, refer to the config() function.
+"""
+import os
+
+import padertorch
```
jensheit, line 12: we are trying to establish the standard to use import padertorch as pt

padertorch/contrib/examples/vae_mnist/mnist_vae_example.py
```diff
@@ -0,0 +1,195 @@
...
+            std=std
+        )
+
+    def review(self, inputs, outputs):
+        images = rearrange(inputs[0], 'b c x y -> b (c x y)')
+        prediction = rearrange(outputs['prediction'], 'b c x y -> b (c x y)')
+
+        MSE = self.MSE_loss(prediction, images)  # reconstruction error
+        # Kullback-Leibler divergence
+        KLD = -0.5 * torch.sum(1 + outputs['std'] - outputs['mean'].pow(2) - outputs['std'].exp())
```
jensheit, line 77: there is already a preimplemented version of the KL-divergence in padertorch.ops.losses

padertorch/contrib/examples/vae_mnist/mnist_vae_example.py
```diff
@@ -0,0 +1,195 @@
...
+        return dict(
+            loss=loss,
+        )
+
+
+class VAE(nn.Module):
+    """ A simple Variational Autoencoder"""
+
+    def __init__(self):
+        super(VAE, self).__init__()
```
jensheit, line 90: this can be replace by super().__init__()

padertorch/contrib/examples/vae_mnist/mnist_vae_example.py
```diff
@@ -0,0 +1,195 @@
...
+                                                   stride=2)
+        self.relu = torch.nn.ReLU()
+
+    def encode(self, x):
+        x = self.conv_1(x)
+        x = self.lrelu(x)
+        x = self.conv_2(x)
+        x = self.lrelu(x)
+        x = self.conv_3(x)
+        x = self.lrelu(x)
```
jensheit, line 121: you can replace the 6 lines above with one line using nn.Sequential

padertorch/contrib/examples/vae_mnist/mnist_vae_example.py
```diff
@@ -0,0 +1,195 @@
...
+    def decode(self, x):
+        x = self.linear_3(x)
+        x = rearrange(x, 'b (x y) -> b x y', y=2, x=2)
+        x = x[:, None, :, :]
+        x = self.transconv_1(x)
+        x = self.relu(x)
+        x = self.transconv_2(x)
+        x = self.relu(x)
+        x = self.transconv_3(x)
+        return x
```
jensheit, line 137: you can replace the 6 lines above with one line using nn.Sequential

padertorch/contrib/examples/vae_mnist/mnist_vae_example.py
```diff
@@ -0,0 +1,195 @@
...
+        y = self.decode(x)
+        return y[:, :, :28, :28], mu, std
+
+
+@ex.config
+def config():
+    epochs = 2          # Use resume=True to train for more epochs
+    batch_size = 128
+    resume = False      # PT: Continue from checkpoints
+    logfilepath = os.environ['STORAGE_ROOT'] + '/mnist'  # PT creates log files
```
jensheit, line 150: use should at least allow for some options in the VAE to be changed here

jensheit, line 150: you should first check whether the STORAGE_ROOT variable is set and give a more informative error message

padertorch/contrib/examples/vae_mnist/mnist_vae_example.py
```diff
@@ -0,0 +1,195 @@
...
+        stop_trigger=(epochs, 'epoch'),
+        #checkpoint_trigger=(5000, 'iteration'),
+        summary_trigger=(1000, 'iteration'),
+    )
+    trainer.register_validation_hook(validation_iterator=testloader)
+    try:
+        trainer.train(trainloader, resume=resume)
+    except Exception:
+        print('#' * 1000)
+        raise
```
jensheit, line 195: why are you doing this try exception catch?

padertorch/contrib/examples/vae_mnist/mnist_vae_example.py
```diff
@@ -9,58 +9,62 @@
...
 import sacred
 
 import torch
 import torch.nn as nn
-import torchvision
-import torchvision.transforms as transforms
 
 from einops import rearrange
 
+from padertorch.contrib.je.data.transforms import Collate
```
jensheit, line 22: You should try not to be dependent on functions from contrib, better use pt.data.collate_fn

padertorch/contrib/examples/vae_mnist/mnist_vae_example.py
```diff
@@ -86,39 +94,38 @@ def review(self, inputs, outputs):
 class VAE(nn.Module):
     """ A simple Variational Autoencoder"""
 
-    def __init__(self):
-        super(VAE, self).__init__()
-        self.kernel_size = 4
-
+    def __init__(self, activation_function, hidden_channels):
+        super().__init__()
+        kernel_size = 4
```
jensheit, line 99: why is the kernel size not an input to the init?

padertorch/contrib/examples/vae_mnist/mnist_vae_example.py
```diff
@@ -145,13 +147,18 @@ def forward(self, input):
 @ex.config
 def config():
     epochs = 2          # Use resume=True to train for more epochs
-    batch_size = 128
+    batch_size = 10
     resume = False      # PT: Continue from checkpoints
+
+    assert os.environ.get('STORAGE_ROOT') is not None, 'STORAGE_ROOT is not defined in the environment variables'
     logfilepath = os.environ['STORAGE_ROOT'] + '/mnist'  # PT creates log files
```
jensheit, line 154: I would choose a storage_dir name which at least gives some indication of the performed task

padertorch/contrib/examples/vae_mnist/mnist_vae_example.py
FrederikRautenberg: I have used the function pt.data.utils.collate_fn. But i get a list of Tensors here, so that i can put them into my net, i had to edit them like this:  " torch.tensor([tensor.numpy() for tensor in inputs['image']])[:, None, :, :] "this was with Collate not necessary. Maybe you can help me to find a better solution.

padertorch/contrib/examples/vae_mnist/mnist_vae_example.py
FrederikRautenberg: should I add an additional condition for the kernel size here? with three layers, the kernel size cannot be arbitrarily large

jensheit: You could add a maximum size, however why shouldnt one use a kernel size of 3 or 5?

padertorch/contrib/examples/vae_mnist/mnist_vae_example.py
```diff
@@ -0,0 +1,216 @@
...
+        images = rearrange(outputs['input'], 'b c x y -> b (c x y)')
+        prediction = rearrange(outputs['prediction'], 'b c x y -> b (c x y)')
+
+        mse = nn.functional.mse_loss(prediction, images)  # reconstruction error
+
+        q = Normal(loc=outputs['mean'], scale=outputs['std'])  # posterior distribution
+        mean = torch.zeros(outputs['mean'].shape[-1])
+        std = torch.ones(outputs['std'].shape[-1])
+        p = Normal(loc=mean, scale=std)  # prior distribution zero mean and unit variance
+        kld = gaussian_kl_divergence(q=q, p=p)  # regularisation error
```
JanekEbb, line 94: In the case of a standard normal prior I would personally prefer the direct computation of the kld as it is only one line of code: 
`kld = -0.5 * (1 + log_var - mu.pow(2) - log_var.exp()).sum()`

padertorch/contrib/examples/vae_mnist/mnist_vae_example.py
```diff
@@ -0,0 +1,216 @@
...
+            torch.nn.ConvTranspose2d(in_channels=hidden_channels, out_channels=1, kernel_size=kernel_size,
+                                     stride=stride),
+        )
+        self.linear_3 = torch.nn.Linear(in_features=1, out_features=self.size_bottleneck**2)
+
+    def encode(self, x):
+        x = self.conv_encoder(x)
+        x = rearrange(x, 'b c x y -> b (c x y)')
+        mu = self.linear_1(x)  # mean value
+        std = 0.5 * self.linear_2(x)  # standard deviation
```
JanekEbb, line 149: Is there a reason for the factor 0.5 here? Also I think it is more common to output `log_var=self.linear_2(x)` with `std = torch.exp(0.5*log_var)`

padertorch/contrib/examples/vae_mnist/mnist_vae_example.py
```diff
@@ -0,0 +1,216 @@
...
+            activation_function(),
+            torch.nn.Conv2d(in_channels=hidden_channels, out_channels=hidden_channels, kernel_size=kernel_size,
+                            stride=stride, padding=self.padding),
+            activation_function(),
+            torch.nn.Conv2d(in_channels=hidden_channels, out_channels=1, kernel_size=kernel_size, stride=stride,
+                            padding=self.padding),
+            activation_function(),
+        )
+        self.linear_1 = torch.nn.Linear(in_features=self.size_bottleneck**2, out_features=1)
+        self.linear_2 = torch.nn.Linear(in_features=self.size_bottleneck**2, out_features=1)
```
JanekEbb, line 130: I think the naming is weird. I would have expected that size_bottleneck would be the size of the latent embedding. However, if I understand correctly, the latent dim is one here (which actually is a much smaller bottleneck).

padertorch/contrib/examples/vae_mnist/mnist_vae_example.py
```diff
@@ -0,0 +1,216 @@
...
+        mean = torch.zeros(outputs['mean'].shape[-1])
+        std = torch.ones(outputs['std'].shape[-1])
+        p = Normal(loc=mean, scale=std)  # prior distribution zero mean and unit variance
+        kld = gaussian_kl_divergence(q=q, p=p)  # regularisation error
+
+        loss = mse + torch.sum(kld)
+
+        return dict(
+            loss=loss,
+        )
```
JanekEbb, line 100: it would be nice if some images would be added (originals and reconstructions) to the review which can be viewed in the tensorboard.

padertorch/contrib/examples/vae_mnist/mnist_vae_example.py
```diff
@@ -0,0 +1,216 @@
...
+        model=model,
+        storage_dir=logfilepath,
+        optimizer=pt_opt.SGD(),
+        stop_trigger=(epochs, 'epoch'),
+        #checkpoint_trigger=(5000, 'iteration'),
+        summary_trigger=(1000, 'iteration'),
+    )
+
+    trainer.register_validation_hook(validation_iterator=test_set)
+    trainer.train(train_set, resume=resume)
```
JanekEbb, line 216: To show that the model is actually working, it would be cool if you finally generate some new images by decoding some z's sampled from a standard normal distribution and save them in logfilepath.

padertorch/contrib/examples/vae_mnist/mnist_vae_example.py
```diff
@@ -0,0 +1,216 @@
...
+@ex.config
+def config():
+    epochs = 2          # Use resume=True to train for more epochs
+    batch_size = 10
+    resume = False      # PT: Continue from checkpoints
+
+    assert os.environ.get('STORAGE_ROOT') is not None, 'STORAGE_ROOT is not defined in the environment variables'
+    logfilepath = os.environ['STORAGE_ROOT'] + '/vae_mnist'  # PT creates log files
+
+    activation_function = ACTIVATION_FN_MAP['relu']  # activation function after convolutional layers
```
JanekEbb, line 174: The mapping shouldn't be called within config

padertorch/contrib/examples/vae_mnist/mnist_vae_example.py
```diff
@@ -0,0 +1,216 @@
...
+
+
+@ex.config
+def config():
+    epochs = 2          # Use resume=True to train for more epochs
+    batch_size = 10
+    resume = False      # PT: Continue from checkpoints
+
+    assert os.environ.get('STORAGE_ROOT') is not None, 'STORAGE_ROOT is not defined in the environment variables'
+    logfilepath = os.environ['STORAGE_ROOT'] + '/vae_mnist'  # PT creates log files
```
JanekEbb, line 172: I would personally prefer that each new experiment uses a new subdirectory in `os.environ['STORAGE_ROOT'] + '/vae_mnist'`

padertorch/contrib/examples/vae_mnist/mnist_vae_example.py
jensheit: We could add those images to a README.md to show what the user can expect when training this model

padertorch/contrib/examples/vae_mnist/mnist_vae_example.py
jensheit: This should be adapted to our new storage root behavior as showcased in the other examples.

PR #24, opened 2020-06-17
[WIP] add multi gpu support

boeddeker:
- Use the virtual mini batch size to distribute to different GPUs
 - Should work with most models without modification
 - Refactor the training loop to support multiple GPUs and
   - The iteration of the trainer can now be manipulated from hooks (e.g. back off)
   - The iteration indicates now the number of optimizer steps and not the number of example used
 - The step function signature changes to support different devices
 - The optimizer review is now separately reported to the hooks
 - Changed _maybe_add_loss_to_review to _review_to_loss_and_summary
 - Add log_error_state for better debugging
 - Remove Trainer.backward
 - Dropped backward in train_step
 - Add gradient check

 - rename train_iterator to train_dataset

 - [x] Support multiple GPUs to process examples from the virtual mini batch
 - [ ] Support multiple GPUs for validation

### DIFF
--- a/padertorch/contrib/examples/multi_gpu/train.py
+++ b/padertorch/contrib/examples/multi_gpu/train.py
@@ -1,0 +1,158 @@
+"""
+A basic example to demonstrate, how DataParallel can be used in padertorch.
+We recommend to use the build in support for data parallel, because it uses the
+virtual mini batch instead of the real mini batch to distribute the work on
+multiple GPUs.
+
+The code is inspired from the examples in
+https://pytorch.org/tutorials/beginner/blitz/data_parallel_tutorial.html
+Authors: Sung Kim and Jenny Kang
+
+Use the following command:
+
+    python train.py with storage_root=/path/to/dir
+
+to start the experiment an take a look at the
+
+    forward shape: torch.Size([11, 28])
+    forward shape: torch.Size([12, 28])
+    review shape: torch.Size([23, 28])
+
+in the output. You should see, that the shape in the forward is different
+from the shape in the review. Also the forward is more offten called than the
+review.
+"""
+from pathlib import Path
+import numpy as np
+
+import torch
+import torch.nn
+
+import sacred
+import sacred.commands
+
+import paderbox as pb
+import padertorch as pt
+from padertorch.modules import fully_connected_stack
+
+from padertorch.contrib.cb.io import get_new_folder
+
+experiment = sacred.Experiment()
+
+
+class MyModel(pt.base.Model):
+
+    def __init__(self, net):
+        super().__init__()
+        self.net = net
+
+    def forward(self, example):
+        print("\tforward shape:", example['feature'].size())
+        out = self.net(example['feature'])
+        return out
+
+    def review(self, example, prediction):
+        print("\treview shape:", example['feature'].size())
+        loss = torch.nn.CrossEntropyLoss()(prediction, example['label'])
+
+        with torch.no_grad():
+            _, predicted = torch.max(prediction, 1)
+            acc = (predicted == example['label']).sum().item() / len(example['label'])
+
+        return dict(loss=loss, scalars={'Accuracy': acc})
+
+
+@experiment.config
+def config():
+    epochs = 2
+    storage_root = None
+
+    features = 28
+    classes = 10
+    
+    trainer = pb.utils.nested.deflatten({
+        'model.factory': MyModel,
+        'model.net.factory': fully_connected_stack,
+        'model.net.input_size': features,
+        'model.net.hidden_size': [10],
+        'model.net.output_size': classes,
+        'model.net.activation': 'relu',
+        'model.net.dropout': 0.,
+
+        'optimizer.factory': pt.optimizer.SGD,
+        'summary_trigger': (1, 'epoch'),
+        'checkpoint_trigger': (1, 'epoch'),
+        'stop_trigger': (2, 'epoch'),
+        'storage_dir': str(get_new_folder(storage_root, mkdir=False)),
+    })
+    pt.Trainer.get_config(trainer)
+
+    experiment.observers.append(sacred.observers.FileStorageObserver.create(
+        Path(trainer['storage_dir']) / 'sacred'
+    ))
+
+
+def get_random_dataset(length, features=28, classes=10):
+    """
+    >>> from paderbox.utils.pretty import pprint
+    >>> np.random.seed(0)
+    >>> pprint(get_random_dataset(10)[0], max_array_length=10)
+    {'feature': array(shape=(25, 28), dtype=float32),
+     'label': array(shape=(25,), dtype=int64)}
+    """
+
+    dataset = []
+    for i in range(length):
+        frames = np.random.randint(20, 30)
+        label = np.random.randint(0, classes)
+        label = np.array([label] * frames, dtype=np.int64)
+        
+        dataset.append({
+            'feature': (label[:, None] + np.random.normal(size=(frames, features))).astype(np.float32),
+            'label': label,
+        })
+    
+    return dataset
+
+
+class DataParallel(torch.nn.DataParallel):
+    def __getattr__(self, attr):
+        """
+        Overwrite __getattr__ so the user does not recognize if the
+        model is wrapped by DataParallel, i.e. forward attribute access
+        to the model.
+        """
+        try:
+            # torch.nn.module overwrites __getattr__,
+            # hence first call super.
+            return super().__getattr__(attr)
+        except AttributeError:
+            pass
+        return getattr(self.module, attr)
+
+
+@experiment.automain
+def main(_run, _config, features, classes):
+    # Get Datasets
+    train_ds = get_random_dataset(10, features=features, classes=classes)
+    dev_ds = get_random_dataset(1, features=features, classes=classes)
+
+    sacred.commands.print_config(_run)
+    trainer: pt.Trainer = pt.Trainer.from_config(_config['trainer'])
+
+    pb.io.dump(_config, trainer.storage_dir / 'config.yaml')
+    pb.io.dump(_config, trainer.storage_dir / 'config.json')
+
+    if torch.cuda.device_count() > 1:
+        print("Let's use", torch.cuda.device_count(), "GPUs!")
+        # dim = 0 [30, xxx] -> [10, ...], [10, ...], [10, ...] on 3 GPUs
+        trainer.model = DataParallel(trainer.model)
+    else:
+        print('#' * 79)
+        print('WARNING: Could not find multiple GPUs !!!')
+        print('#' * 79)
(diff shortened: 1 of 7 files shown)


padertorch/contrib/examples/tasnet/train.py
```diff
@@ -12,6 +12,8 @@
 from sacred import Experiment
 import sacred.commands
 
+# sacred.SETTINGS.CONFIG.READ_ONLY_CONFIG = False
+
```
jensheit, line 16: Is this comment necessary?

padertorch/train/trainer.py
```diff
@@ -248,7 +249,16 @@ def train(
         # Change model to train mode (e.g. activate dropout)
         self.model.train()
 
-        self.to(device)
+        if isinstance(device, (tuple, list)) \
+                and all([isinstance(d, int) for d in device]):
+            # multiple devises e.g. [0, 1], [0, 1, 2, 3], ...
```
jensheit, line 254: the second if statement should be an assert in my opinion

padertorch/train/trainer.py
```diff
@@ -266,48 +276,111 @@ def train(
...
+                            # torch.nn.parallel.DataParallel.
+                            # We also use the same functions
+                            # (i.e. replicate, parallel_apply and gather).
+                            #
+                            # The difference is, that we need no scatter,
+                            # because we simply use multiple examples and
+                            # the gather must only be applied on the loss.
+
+                            # Move copies of the model to each GPU
+                            replicas = replicate(self.model, device[:len(example)])
```
jensheit, line 342: should we not do this before the virtual minibatch loop because now we are copying the same model to the devices in each iteration

padertorch/train/trainer.py
boeddeker: Yes, an assert produces a better exception msg. I will change it

padertorch/train/trainer.py
boeddeker: I copied the pattern from torch.nn.parallel.DataParallel.
They do the replicate in each forward call.
I haven't measured the overhead.
To keep the code simple, I would keep it for the moment here.

jensheit: I think it makes sense to copy it in each forward call if you also always do a backwards step because the model is changing, but in our case we are just doing a backward step once for multiple forwards so we could copy it once

padertorch/contrib/examples/tasnet/train.py
boeddeker: Not really. I added it, because I expected that the script fails without disable `READ_ONLY_CONFIG`.
Someone reported that the `betas` in the Adam optimizer caused some problems.
But this may only fail in some pytorch versions.

I will activate that line, to prevent some strange error messages from sacred.

boeddeker:
I want to merge this PR and do the open points later:
- Support data parallel in validate
- Create the replicates only once after optimize
- Figure out which torch version have a working DataParallel (e.g. 1.4 doesn't work with RNNs)

Any comments to this PR?

I changed also the profiling results.
It is now simpler. The time per iteration is now measures explicit and all relative values are relative to the iteration time.

padertorch/contrib/examples/multi_gpu/train.py
```diff
@@ -0,0 +1,164 @@
...
+        trainer.model = DataParallel(trainer.model)
+    else:
+        print('#' * 79)
+        print('WARNING: Could not find multiple GPUs !!!')
+        print('#' * 79)
+    print('torch.cuda.device_count()', torch.cuda.device_count())
+    print('torch.cuda.is_available()', torch.cuda.is_available())
+
+    trainer.register_validation_hook(validation_iterator=dev_ds)
+    trainer.train(train_ds)
```
jensheit, line 164: Shouldn't we add an example on how to use multiple gpus here instead of one showing what not to do?

padertorch/train/trainer.py
```diff
@@ -362,39 +469,65 @@ def optimizer_zero_grad(self):
...
-    def validation_step(self, example):
-        return self.step(example, self.validate_timer)
+    def validation_step(self, model, example, device):
+        # [1:] -> ignore the loss. Is already in scalars.
+        return self.step(model, example, self.validate_timer, device)[1:]
 
-    def step(self, example, timer):
+    def step(self, model, example, timer, device):
         # TODO: Backup OutOfMemory
+        # with self.train_timer['time_per_step']:
```
thequilo, line 502: Time per step is no longer reported?

padertorch/train/trainer.py
```diff
@@ -362,39 +469,65 @@ def optimizer_zero_grad(self):
...
         with timer['time_per_to_device']:
-            example = self.model.example_to_device(example, self.device)
+            example = self.model.example_to_device(
+                example, device
+            )
```
thequilo, line 506: Why reformatted? The original line is shorter than 79 chars.

thequilo: Should this be `model` instead of `self.model`?

padertorch/contrib/examples/multi_gpu/train.py
boeddeker: Good point.
I created first this example, before I thought a build in support would be better,
I will change it to show both options.

padertorch/train/trainer.py
```diff
@@ -257,57 +267,148 @@ def train(
...
+                            new_epoch = False
+                        elif minibatch_index == 0:
+                            # Call pre_step after getting the next example,
+                            # to correctly detect the next epoch
+                            with timer.pause():
+                                for hook in hooks:
+                                    hook.pre_step(self)
+
+                        if len(device) == 1:
+                            example, = example
```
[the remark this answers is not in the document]
boeddeker: This should never fail, this is just for the case that you change the trainer code.
But I will change the code to be more verbose (i.e. use your suggestion).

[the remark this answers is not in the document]
thequilo: Yes, it "should" never fail. But you never know, maybe you at some point change the code before this line and it suddenly can fail

padertorch/train/trainer.py
```diff
@@ -257,57 +267,148 @@ def train(
...
-                        )
+            train_iterable = None
+            while True:
+                new_epoch = False
+                if train_iterable is None:
+                    new_epoch = True
 
+                    # Call pre_step between the epochs.
+                    # We call it here, so it is done, before the iteration
+                    # over the train_dataset starts.
```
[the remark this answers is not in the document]
boeddeker: In general, it is not necessary.
The reason is, that this can reduce the memory footprint.
When you trigger the validation after an epoch the train dataset shouldn't start to load the next examples.

padertorch/train/trainer.py
```diff
@@ -257,57 +267,148 @@ def train(
...
+                            # Release pytorch object to reduce memory footprint
+                            del example
+                            del model_output
+                            del review
+
+                            with self.train_timer['time_per_backward']:
+                                loss.backward(retain_graph=False)
+                            del loss
+
+                    # Only the summary hook will use optimizer_review
```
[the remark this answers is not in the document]
boeddeker: At the moment: yes.
I am also not completely happy with this solution.
Do you have another idea, how we can report the values from the optimizer?

padertorch/train/trainer.py
```diff
@@ -257,57 +267,148 @@ def train(
...
+                    if optimize:
+                        with self.train_timer['time_per_optimize']:
+                            optimizer_review = self.optimizer_step()
+                            for hook in hooks:
+                                hook.post_step(self, None, None, optimizer_review)
```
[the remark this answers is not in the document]
boeddeker: Good idea. @JanekEbb and @jensheit what do you think?

padertorch/train/trainer.py
boeddeker: I dropped it, because it is relative similar to `time_per_iteration` ( = `time_per_data_loading` + `time_per_step`). Since backward is no longer in the step, it cannot directly be measured.

padertorch/train/trainer.py
boeddeker: Reformatted: This was in the merge, I didn't carefully check and used the previous formatting before you changed it. I will change it back.

model: It doesn't matter, but using `model` is better.

padertorch/train/trainer.py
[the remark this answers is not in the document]
jensheit: sounds good to me, and the default would be that post_optimize just calls post_step?

padertorch/train/trainer.py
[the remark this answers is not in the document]
boeddeker: As far as I know is `example, = example` a very cheap operation and if someone changes the code in the Trainer it is ok, to get a less informative error message (Better than silently fail).
But to be more verbose, I will add an assert.

padertorch/train/trainer.py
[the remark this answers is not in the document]
boeddeker: I think the default should be that `post_optimize` does nothing.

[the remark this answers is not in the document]
jensheit: If post_optimize does nothing then we have to add the behavior to the hooks which depent on the optimizer. In most cases this means calling the post_step or implement a similar behavior.

boeddeker:
I have now a first number for the speedup of the DPRNN.
With 1 GPU the time per iteration is 0.74 and with 2 GPUs it is 0.4.
I think such a speedup is ok.

The replicate (i.e. copy the model to the second GPU) takes 3% of the execution time.
I think that is also ok.

padertorch/train/trainer.py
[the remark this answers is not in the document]
jensheit: but you are right, the lines of code we are saving are not worth the possible confusing it may cause

jensheit:
> I have now a first number for the speedup of the DPRNN.
> With 1 GPU the time per iteration is 0.74 and with 2 GPUs it is 0.4.
> I think such a speedup is ok.
> 
> The replicate (i.e. copy the model to the second GPU) takes 3% of the execution time.
> I think that is also ok.

Great, yes that looks like it is working.

padertorch/train/trainer.py
boeddeker: done

padertorch/train/trainer.py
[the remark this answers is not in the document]
boeddeker: I pushed now a suggestion for `post_optimize`

padertorch/train/trainer.py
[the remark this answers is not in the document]
boeddeker: done

padertorch/train/trainer.py
boeddeker: done

padertorch/train/trainer.py
boeddeker: I will keep this point for another PR.

PR #26, opened 2020-06-24
doc: Virual minibatch size and multiple GPUs

boeddeker:
I made a proposal for documenting multiple GPUs.

What do you think? (I didn't check for typos)

### DIFF
--- a/doc/virtual_batch_size_multi_gpu.md
+++ b/doc/virtual_batch_size_multi_gpu.md
@@ -1,6 +1,6 @@
 # Virual minibatch size and multiple GPUs
 
-Note: The code examples here should be interpreted as pseudo code, i.e. they should show what the code does, the actual implementation is more complicated, because it does more things.
+Note: The code examples here should be interpreted as pseudo code, i.e. they should show what the code does. The actual implementation is more complicated, because it does more things.
 
 ## What is (Virual) minibatch size?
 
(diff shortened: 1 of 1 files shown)


jensheit:
Perhaps you should introduce the name virtual minibatch earlier to clarify from the beginning that both uses of a minibatch are not equivalent. It is a little confusing that both use cases are called minibatch.

doc/virtual_batch_size_multi_gpu.md
```diff
@@ -0,0 +1,117 @@
...
+    if i == virtual_minibatch_size:                 # <-----
+        i = 0                                       # <-----
+        optimizer.step()                            # <-----
+        optimizer.zero_grad()
+```
+
+Both options for the minibatch size can be combined.
+The effective minibatch size for the optimizer will be the dataset minibatch size times the `virtual_minibatch_size`. For operations like batch normalization (i.e. operations that work in the minibatch axis) the batch size will be the minibatch that is used to produce the dataset, NOT the virtual batch size.
+
+## Why use (Virtual) minibatch size?
```
thequilo, line 48: Mention that it is simliar to `accumulate_grads` (I think) from ESPnet?

thequilo: It's called `accum_grad` in ESPnet

jensheit:
> How do we handle proof-reading for grammar/typos/punctuation? I think it's not a good idea to add a comment for every typo.

Just directly commit the changes you want to do.

doc/virtual_batch_size_multi_gpu.md
boeddeker: Du you want to add this the comparison with ESPnet?

boeddeker:
Can we merge?

PR #28, opened 2020-06-25
fix summary buffer and annealing hooks

JanekEbb:
- fix summary buffer extension by adding to_list(...)
- fix annealing hooks to work with backoff
- add NotTrigger to allow lr annealing (rampup) jointly with backoff without conflicts

### DIFF
--- a/padertorch/train/hooks.py
+++ b/padertorch/train/hooks.py
@@ -223,7 +223,7 @@ class SummaryHook(TriggeredHook):
             self.summary['histograms'][key] = \
                 self.summary['histograms'][key][-1000000:]
         for key, buffer in popped_review.pop('buffers', dict()).items():
-            self.summary['buffers'][key].extend(self._detach(buffer))
+            self.summary['buffers'][key].extend(self._to_list(self._detach(buffer)))
         for key, snapshot in popped_review.pop('snapshots', dict()).items():
             self.summary['snapshots'][key] = self._detach(snapshot)  # snapshot
         for key, audio in popped_review.pop('audios', dict()).items():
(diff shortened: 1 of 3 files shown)


padertorch/train/hooks.py
```diff
@@ -223,7 +223,7 @@ def update_summary(self, review):
             self.summary['histograms'][key] = \
                 self.summary['histograms'][key][-1000000:]
         for key, buffer in popped_review.pop('buffers', dict()).items():
-            self.summary['buffers'][key].extend(self._detach(buffer))
+            self.summary['buffers'][key].extend(self._to_list(self._detach(buffer)))
```
boeddeker, line 226: @TCord Why is here `extend` used? I would say it should be `append`. We should make no assumptions about the buffer and don't manipulate the value (except detach). So don't call `_to_list`. For other entries `_to_list` is fine, because the value will be prepared for tensorboardX

padertorch/train/hooks.py
```diff
@@ -673,13 +673,13 @@ def _back_off(self, trainer: 'pt.Trainer'):
         latest_symlink_path.symlink_to(best_ckpt)
 
         best_iter = int(best_ckpt[len('ckpt_'): -len('.pth')])
-        for j in range(1, len(self.ckpt_ranking)):
-            ckpt = self.ckpt_ranking[-j][0]
+        for j in range(len(self.ckpt_ranking)-1, 0):
```
boeddeker, line 676: Are you sure, that this does what you want?
```python
>>> l = [1, 2, 3]
>>> list(range(len(l)-1, 0))
[]
```

padertorch/train/hooks.py
TCord: I used `extend` to make it similar to the other summary types, but forgot that it inherently makes assumptions on the data again. I also think, `append` would be the better choice here.

padertorch/train/hooks.py
JanekEbb: Fixed it and added a test case

padertorch/train/hooks.py
JanekEbb: changed it

PR #34, opened 2020-06-30
add stft to padertorch ops

jensheit:
This is an STFT implementation for torch which allows for close to perfect reconstruction of the original time signal by using STFT.inverse. 
Most of the code is inspired by paderbox.transform.stft

A similar code was used in [Demystifying TasNet: A Dissecting Approach ](https://arxiv.org/abs/1911.08895)


padertorch/ops/stft_module.py
```diff
@@ -0,0 +1,270 @@
+import numpy as np
```
thequilo, line 1: Why is this file called `stft_module`? This doesn't match the other files in `padertorch/ops`. As long as we only have an `STFT` class, there's no name clash

padertorch/ops/stft_module.py
```diff
@@ -0,0 +1,270 @@
...
+        Args:
+
+            size: Scalar FFT-size.
+            shift: Scalar FFT-shift. Typically shift is a fraction of size.
+            window: Window function handle.
+            window_length:
+            fading:
+            pad:
+            symmetric_window:
+            complex_representation: str, defines how to handle the real and
```
thequilo, line 77: `str` is redundant here

padertorch/ops/stft_module.py
```diff
@@ -0,0 +1,270 @@
...
+
+        encoded = encoded.view(*org_shape[:-1], *encoded.shape[-2:])
+        encoded = rearrange(encoded, '... feat frames -> ... frames feat')
+        encoded = torch.chunk(encoded, 2, dim=-1)
+        if self.complex_representation == 'stacked':
+            encoded = torch.stack(encoded, dim=-1)
+        elif self.complex_representation == 'concat':
+            encoded = torch.cat(encoded, dim=-1)
+        else:
+            raise NotImplementedError
```
thequilo, line 166: Is this good code style? I honestly don't know if it is correct to raise the `NotImplementedError` class. I frequently see this, but my intuition says that it should be `NotImplementedError()`. I also prefer it if the error message has some explanatory text. Even if the Exception should never be raised in this code, you can never know.

Btw, for `complex_representation != 'complex'`, a `ValueError` might be more appropriate.

padertorch/ops/stft_module.py
jensheit: I used this name to allow the addition of a stft function. But this is of course open for discussion. If most of you do not think we need the stft function or prefer an other name we can change it.

padertorch/ops/stft_module.py
jensheit: This error should never occour because a new complex_representation should already be catched in the init. This can only happen if somebody overwrites ```STFT.complex_representation``` after initialization. However, I will add a sentence as explanation to the error message.

thequilo: Yes, I know that this "should never happen". But you never know what the user will be doing...

padertorch/ops/stft_module.py
boeddeker: I am also in favor to keep the name `stft` free, so that we can implement the function as alternative to the class.

padertorch/ops/stft_module.py
boeddeker: I think it should be at least `NotImplementedError(self.complex_representation)`. Just in case, that this exception is thrown  the user gets the necessary information about the internal state.

padertorch/ops/stft_module.py
```diff
@@ -0,0 +1,270 @@
...
+            symmetric_window:
+            complex_representation: str, defines how to handle the real and
+                imaginary part of the complex stft signal:
+                                either complex, concat or stacked
+                                complex is not supported at the moment
+        """
+        possible_out_types = ['complex', 'concat', 'stacked']
+        assert complex_representation in possible_out_types, (
+            f'Please choose one of the predefined output_types'
+            f' {possible_out_types}'
```
boeddeker, line 85: Can you add `complex_representation` in the assertion msg?

padertorch/ops/stft_module.py
```diff
@@ -0,0 +1,270 @@
...
+            elif stride != 1 and (x.shape[-1] + stride - length) % stride != 0:
+                pad_size = stride - ((x.shape[-1] + stride - length) % stride)
+                x = F.pad(x, (0, pad_size))
+
+        x = torch.unsqueeze(x, 1) # [..., 1, T]
+        weights = self.stft_kernel.to(x)
+        encoded = F.conv1d(x, weight=weights, stride=stride)
+
+        encoded = encoded.view(*org_shape[:-1], *encoded.shape[-2:])
+        encoded = rearrange(encoded, '... feat frames -> ... frames feat')
```
boeddeker, line 159: Can you do the `view` `rearrange` in a single `rearrange`?
Something like `rearrange(encoded, '... (feat frames) -> ... frames feat', feat=..., frames=...)` should work.

padertorch/ops/stft_module.py
```diff
@@ -0,0 +1,270 @@
...
+            signal_real, signal_imag = torch.chunk(stft_signal, 2, dim=-1)
+        elif self.complex_representation == 'concat':
+            signal_real, signal_imag = torch.chunk(x, 2, dim=-2)
+        else:
+            raise NotImplementedError
+
+        signal_real = torch.cat(
+            [signal_real, signal_real[:, 1:-1].flip(1)], dim=1)
+        signal_imag = torch.cat(
+            [signal_imag, -signal_imag[:, 1:-1].flip(1)], dim=1)
```
boeddeker, line 211: Does this assume that stft size is even?

jensheit: yes, you are right either I have to change it to allow uneven sizes as well or add an assert

boeddeker: I think an assert here would be enough. We can add support for uneven stft sizes when we find an use case.

padertorch/ops/stft_module.py
jensheit: done

padertorch/ops/stft_module.py
jensheit: done

padertorch/ops/stft_module.py
jensheit: done

padertorch/ops/stft_module.py
jensheit: done

padertorch/ops/stft_module.py
jensheit: okay so should we keep the name or change it to something like transform?

thequilo: I also think that we should keep `stft` free for a functional implementation. But I also don't like the suffix `_module` because it is not used anywhere else (in `padertorch`). 

I just noticed that `einsum` module contains a function `einsum`. So, I think this should be similar in all sub-modules of `pt.ops`, so it should either be `stft` with everything exported to `__init__` (as in `einsum`), or all other modules have to be renamed as well.

jensheit: changed this to _stft.py

PR #35, opened 2020-07-08
Example for modular tasnet

jensheit:
This example for tasnet allows to switch between the learnable encoder and decoder and the stft / istft. In future a ConvNet separator could be added to allow the user to switch between the DPRNN and ConvNet Separator.

The results achieved with this example are similar to the ones presented in [Demystifying TasNet: A Dissecting Approach](https://arxiv.org/abs/1911.08895)

### DIFF
--- a/padertorch/contrib/examples/tasnet/tas_coders.py
+++ b/padertorch/contrib/examples/tasnet/tas_coders.py
@@ -172,21 +172,21 @@ class StftEncoder(torch.nn.Module):
             fading=False, complex_representation='concat'
         )
 
-    def forward(self, inputs, num_samples=None):
+    def forward(self, inputs, sequence_lengths: torch.Tensor = None
+    ) -> Tuple[torch.Tensor, Union[torch.Tensor, None]]:
         """
         Args:
             inputs: shape: [..., T], T is #samples
-            num_samples: list or tensor of #samples
+            sequence_lengths: list or tensor of #samples
         Returns:
             [..., N, frames] the stft encoded signal
         """
 
         encoded = self.stft(inputs)
         encoded = rearrange(encoded, '... frames fbins -> ... fbins frames')
-        if num_samples is not None:
-
+        if sequence_lengths is not None:
             num_frames = torch.tensor([self.stft.samples_to_frames(samples)
-                                       for samples in num_samples])
+                                       for samples in sequence_lengths])
             return encoded, num_frames
         else:
             return encoded
@@ -226,7 +226,7 @@ class IstftDecoder(torch.nn.Module):
             shift=stride, fading=False, complex_representation='concat'
         )
 
-    def forward(self, stft_signal):
+    def forward(self, stft_signal) -> torch.Tensor:
         """
         Args:
             stft_signal: shape: [B, ..., N, Frames]
--- a/padertorch/contrib/examples/tasnet/tas_coders.py
+++ b/padertorch/contrib/examples/tasnet/tas_coders.py
@@ -12,8 +12,8 @@ class TasEncoder(torch.nn.Module):
     be used as a module for feature extraction in other modules/models.
     """
 
-    def __init__(self, L: int = 20, N: int = 256, stride: int = None,
-                 bias: bool = False):
+    def __init__(self, window_length: int = 20, feature_size: int = 256,
+                 stride: int = None, bias: bool = False):
         """
         Args:
             L: The block size in samples (length of the filters).
@@ -27,14 +27,16 @@ class TasEncoder(torch.nn.Module):
         super().__init__()
 
         if stride is None:
-            stride = L // 2
+            stride = window_length // 2
 
-        self.L = L
-        self.N = N
+        self.window_length = window_length
+        self.feature_size = feature_size
         self.stride = stride
 
-        self.encoder_1d = torch.nn.Conv1d(1, N, L, stride=self.stride,
-                                          padding=0, bias=bias)
+        self.encoder_1d = torch.nn.Conv1d(
+            1, feature_size, window_length,
+            stride=self.stride, padding=0, bias=bias
+        )
 
     def forward(
             self,
@@ -68,14 +70,15 @@ class TasEncoder(torch.nn.Module):
         l = x.shape[-1]
         sq_offset = -1
 
-        if l % (self.L // 2) > 0:
-            padding = self.L // 2 - (l % (self.L // 2))
+        if l % (self.window_length // 2) > 0:
+            padding = self.window_length // 2 - (l % (self.window_length // 2))
             x = F.pad(x, (0, padding))
             sq_offset = 0
 
         # Compute new sequence_lengths
         if sequence_lengths is not None:
-            sequence_lengths = sequence_lengths // (self.L // 2) + sq_offset
+            sequence_lengths = sequence_lengths // (
+                    self.window_length // 2) + sq_offset
 
         # Add channel dimension. Results in (B, 1, T)
         x = torch.unsqueeze(x, dim=1)
@@ -91,8 +94,8 @@ class TasDecoder(torch.nn.Module):
     Encapsulates the decoder of the TasNet in a separate module
     """
 
-    def __init__(self, L: int = 20, N: int = 256, stride: int = None,
-                 bias=False):
+    def __init__(self, window_length: int = 20, feature_size: int = 256,
+                 stride: int = None, bias=False):
         """
         The arguments should match the used encoder.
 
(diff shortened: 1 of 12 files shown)


padertorch/contrib/examples/tasnet/tas_coders.py
```diff
@@ -127,3 +129,90 @@ def forward(self, w: torch.Tensor) -> torch.Tensor:
...
+    >>> import numpy as np
+    >>> stft_out = stft(\
+            mixture.numpy(), 256, 10, window_length=20, \
+            fading=None)
+    >>> stft_encoded = np.concatenate(\
+            [np.real(stft_out), np.imag(stft_out)], axis=-1\
+        ).transpose(0,1,3,2)
+    >>> np.testing.assert_allclose(stft_encoded, encoded, atol=1e-5)
+    """
+    def __init__(self, L: int = 20, N: int = 256, stride: int = None):
```
boeddeker, line 153: I don't like `L` and `N`. Do you have a better name for these variables? I think they will now also appear in the config, so a longer name would be better.

padertorch/contrib/examples/tasnet/tas_coders.py
```diff
@@ -127,3 +129,90 @@ def forward(self, w: torch.Tensor) -> torch.Tensor:
...
+        super().__init__()
+        self.L, self.N, self.stride = L, N, stride
+
+        if stride is None:
+            stride = L // 2
+        self.stft = STFT(size=N-2, shift=stride, window_length=L, fading=False,
+                         complex_representation='concat')
```
boeddeker, line 160: Can you document, that you keep two uninformative values?
I think `N-2` is wrong. It may be something like: `(N - 1) // 2 * 2`

padertorch/contrib/examples/tasnet/tasnet.py
```diff
@@ -226,3 +226,54 @@ def review(self, inputs, outputs):
...
+            mask: If `True`, use the output of the NN as a mask in tas domain
+                for separation. Otherwise, use the output directly as an
+                estimation for the separated signals.
+            output_nonlinearity: Nonlinearity applied to the output (right
+                before masking/decoding)
+            num_speakers: The number of speakers/output streams
+            additional_out_size: Size of the additional output. Has no effect if
+                set to 0.
+        """
+        super().__init__()
```
boeddeker, line 257: Does this work? I think `TasNet.__init__()` will complain.

padertorch/contrib/examples/tasnet/tas_coders.py
jensheit: I can change them to window_size and num_feats but in that case they should also be changed for the encoder and decoder

padertorch/contrib/examples/tasnet/tasnet.py
```diff
@@ -226,3 +226,54 @@ def review(self, inputs, outputs):
 
     def flatten_parameters(self) -> None:
         self.dprnn.flatten_parameters()
+
+
+class ModularTasNet(TasNet):
```
boeddeker, line 231: Is the opposite inheritance not better? `TasNet` initialize encoder separator and decoder and forwards them to `ModularTasNet`.

For now I like to keep `TasNet` to make no breaking change. But in the long term, we may deprecate it.

padertorch/contrib/examples/tasnet/tasnet.py
```diff
@@ -226,3 +226,54 @@ def review(self, inputs, outputs):
...
+    def __init__(
```
boeddeker, line 232: What do you think about a `finalize_dogmatic_config`?
In this way it is easier to switch between example configs.

padertorch/contrib/examples/tasnet/tas_coders.py
```diff
@@ -127,3 +129,90 @@ def forward(self, w: torch.Tensor) -> torch.Tensor:
...
+
+    def forward(self, inputs, num_samples=None):
+        """
+        Args:
+            inputs: shape: [..., T], T is #samples
+            num_samples, list or tensor of #samples
```
thequilo, line 166: Typo: missing `:`

padertorch/contrib/examples/tasnet/tas_coders.py
boeddeker: Since encoder and decoder is hidden for the user at the moment, we can change them too (and should do, to keep them consistent).

But as always: No shortcuts in the signature: `num_feats` -> `num_features`

padertorch/contrib/examples/tasnet/tas_coders.py
```diff
@@ -127,3 +129,90 @@ def forward(self, w: torch.Tensor) -> torch.Tensor:
...
+            stft_signal: shape: [B, ..., N, Frames]
+            num_samples, list or tensor of #samples
+        Returns:
+            [B, ..., T]
+        """
+
+        stft_signal = rearrange(
+            stft_signal, '... fbins frames  -> ... frames fbins')
+
+        return self.stft.inverse(stft_signal)
```
thequilo, line 218: Can you insert a newline at the end of the file?

padertorch/contrib/examples/tasnet/tasnet.py
thequilo: I agree with this

padertorch/contrib/examples/tasnet/tasnet.py
jensheit: yes, I though about that but I did not want to use it in an example since it is a little bit hard to follow

padertorch/contrib/examples/tasnet/tasnet.py
jensheit: yes, the idea is to rework this after the workshop

padertorch/contrib/examples/tasnet/tasnet.py
thequilo: You can also make sure that encoder and decoder configs are synchronized in there

padertorch/contrib/examples/tasnet/tasnet.py
thequilo: It certainly will

padertorch/contrib/examples/tasnet/train_modular.py
```diff
@@ -0,0 +1,356 @@
...
+import paderbox as pb
+import numpy as np
+
+from sacred.observers.file_storage import FileStorageObserver
+from lazy_dataset.database import JsonDatabase
+
+from padertorch.contrib.neumann.chunking import RandomChunkSingle
+from padertorch.contrib.ldrude.utils import get_new_folder
+
+nickname = "dprnn"
```
thequilo, line 33: `nickname="tasnet"`?

padertorch/contrib/examples/tasnet/tasnet.py
jensheit: this works because tasnet has defaults for all inputs but yes it should be ```super(TasNet).__init__()```

padertorch/contrib/examples/tasnet/tasnet.py
boeddeker: Do you mean deprecation or changing the inheritance order?

padertorch/contrib/examples/tasnet/tasnet.py
boeddeker: I hate `super(TasNet).__init__()`, because it looks like python 2 and it is an indicator to rewrite some code.
As mentioned in another comment changing the inheritance order can solve this.

padertorch/contrib/examples/tasnet/tas_coders.py
jensheit: I used num_feats since it is used in tasnet and dprnn but I can change that as well

padertorch/contrib/examples/tasnet/tasnet.py
jensheit: deprecation, I can change the inheritance now

padertorch/contrib/examples/tasnet/tas_coders.py
boeddeker: I didn't check the others. Use the name that you think is better.

padertorch/contrib/examples/tasnet/tasnet.py
boeddeker: ok, thanks

padertorch/contrib/examples/tasnet/tas_coders.py
jensheit: N-2 is only wrong if we allow uneven stft sizes, which we do not support at the moment. But I can replace the size calculation and add an assert, that the feature_size is even.

padertorch/contrib/examples/tasnet/tasnet.py
thequilo: `super(TasNet).__init__()` is wrong (as far as I understand). It should be `super(ModularTasNet, self).__init__()` or better `super().__init__()` and if you really want to skip the `__init__` of `TasNet`, you can write `super(TasNet, self).__init__()`. I generally don't think that skipping a constructor in the MRO is good coding style. And also it looks like Python2.

jensheit: I meant the later one but since we decided to change the inheritance order this problem is already addressed.

PR #37, opened 2020-07-20
replace normalization in core to support batches of examples with dif…

JanekEbb:
…ferent sequence lengths


padertorch/modules/normalization.py
```diff
@@ -1,104 +1,337 @@
...
+        if self.gamma is not None:
+            x = x / self.gamma
+        if self.scale:
+            x = torch.sqrt(self.running_var) * x
+        if self.shift:
+            x = x + self.running_mean
+        return x
+
+
+class Normalize(Function):
```
jensheit, line 216: perhaps you should change this to a private class, since the user is supposed to use the Norm class.
The same holds for the normalize function

jensheit: another option would be to change the name to clarifiy that this is just an intermediate class

padertorch/ops/sequence/mask.py
```diff
@@ -0,0 +1,66 @@
+import torch
+
+
+def compute_mask(x, sequence_lengths, batch_axis=0, sequence_axis=1):
+    """
+    This function calculates a mask which indicates the position of non-padded values.
+    It can be used to do subsequent operations only on non-padded values.
```
jensheit, line 7: I have added a small doctext to describe this function

padertorch/modules/normalization.py
```diff
@@ -1,104 +1,337 @@
...
+            assert gamma.dim() == x.dim(), gamma.shape
+            y = y * gamma
+        if beta is not None:
+            assert beta.dim() == x.dim(), beta.shape
+            y = y + beta
+        return y*mask, mean, power, n_values
+
+    @staticmethod
+    def backward(ctx, grad_y, grad_mean, grad_power, _):
+        # equations from https://arxiv.org/abs/1502.03167
```
jensheit, line 253: You should add a small doctext expalining why you are overwriting the backward step

padertorch/modules/normalization.py
```diff
@@ -1,104 +1,337 @@
...
+        if self.training or not self.track_running_stats:
+            y, mean, power, n_values = normalize(
+                x, gamma=self.gamma, beta=self.beta,
+                statistics_axis=self.statistics_axis,
+                batch_axis=self.batch_axis, sequence_axis=self.sequence_axis,
+                sequence_lengths=sequence_lengths, shift=self.shift, scale=self.scale,
+                eps=self.eps
             )
+            if self.track_running_stats:
+                self.num_tracked_values += n_values.data
```
thequilo, line 167: Are these `.data`s here (and in the following code) remainders of pre-PyTorch 0.4 code? If so, it is safer to replace it with `.detach()` (see https://pytorch.org/blog/pytorch-0_4_0-migration-guide/#what-about-data). If not, I would be interested in why you use `.data` here.

padertorch/modules/normalization.py
```diff
@@ -1,104 +1,337 @@
...
+        # compute mask
+        mask = compute_mask(x, sequence_lengths, batch_axis, sequence_axis)
 
-        return mean, norm
+        # compute statistics
+        n_values = mask.sum(dim=statistics_axis, keepdim=True)
+        x = x * mask
+        mean = x.sum(dim=statistics_axis, keepdim=True) / torch.max(n_values, torch.ones_like(n_values))
+        power = (x ** 2).sum(dim=statistics_axis, keepdim=True) / torch.max(n_values, torch.ones_like(n_values))
+        y = x
```
thequilo, line 235: Why `y=x`?

padertorch/modules/normalization.py
thequilo: If I understood this correctly, the `normalize` function can still be used on its own if you don't want it to show up in the Model's repr, don't need any statistics tracking, and don't like the string axis definition.

padertorch/ops/sequence/mask.py
```diff
@@ -0,0 +1,66 @@
...
+             [1., 1., 1., 0.],
+             [1., 1., 1., 0.],
+             [1., 1., 1., 0.],
+             [1., 1., 1., 0.]]])
+
+    Args:
+        x:
+        sequence_lengths:
+        batch_axis:
+        sequence_axis:
```
thequilo, line 48: Can you add an explanation for these or remove the section? I think that the `Args` and `Returns` sections don't make sense if they don't contain a description of what is going on.

thequilo: E.g., it would be interesting to know that it returns a one-mask if `sequence_legnths is None`.

padertorch/modules/normalization.py
jensheit: yes, that is true but as a default people should be encouraged to use the Norm class and only use the normalize function if they know what they are doing

jensheit: and even if you are using the normalize function the Normalize class is still hidden

PR #38, opened 2020-07-27
adds set_last to ProgressbarHook initialisation during training

jensheit:
adds set_last to update the iteration counter of the ProgressbarHook in case of resume

### DIFF
--- a/padertorch/train/trainer.py
+++ b/padertorch/train/trainer.py
@@ -272,6 +272,8 @@ class Trainer(Configurable):
                 # TypeError: object of type '...' has no len()
                 max_it_len = None
             hooks.append(ProgressBarHook(self._stop_trigger, max_it_len))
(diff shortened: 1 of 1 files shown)


padertorch/train/trainer.py
```diff
@@ -272,6 +272,8 @@ def train(
                 # TypeError: object of type '...' has no len()
                 max_it_len = None
             hooks.append(ProgressBarHook(self._stop_trigger, max_it_len))
+            # set_last updates the iteration counter in case of resume
+            hooks[-1].set_last(self.iteration, self.epoch)
```
boeddeker, line 276: Can you make the `set_last` call explicit? (i.e. `progress_bar_hook.set_last(...)`)

jensheit: Done

PR #41, opened 2020-08-10
Move Deep Clustering to contrib

TCord:
Removes BSS models from model directory as models are currently unused. DC example will follow in the future.


TCord:
I think if we have the test case and the model is working we can keep the test cases in the test folder instead of moving them to the contrib folder. It only indicates that the model in contrib is usable, so we don't need to include a test for every example.

boeddeker:
I think when we have a "cheap" test for a model, we can move the tests to the normal test folder, when they run on azure.

PR #45, opened 2020-08-12
split from_config_and_checkpoint

jensheit:
Split Module.from config_and_checkpoint to allow loading checkpoints for an already initialized module with mpi.

### DIFF
--- a/padertorch/base.py
+++ b/padertorch/base.py
@@ -96,7 +96,7 @@ class Module(nn.Module, Configurable, abc.ABC):
             map_location='cpu',
             consider_mpi=False,
     ) -> 'Module':
-        """Instantiate the module from given config and checkpoint.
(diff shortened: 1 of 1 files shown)


padertorch/base.py
```diff
@@ -81,6 +81,39 @@ def from_config_and_checkpoint(
...
+
+    def load_checkpoint(
+            self,
+            checkpoint_path: (Path, str),
+            in_checkpoint_path: str = 'model',
+
+            map_location='cpu',
+            consider_mpi=False,
+    ) -> 'Module':
+        """Instantiate the module from given config and checkpoint.
```
boeddeker, line 99: Can you fix this text?

jensheit: Done

PR #46, opened 2020-08-14
Move get_new_folder to padertorch.io

boeddeker:
- Move `get_new_folder` to `padertorch.io`
    - add `id_type` argument to switch between `index` and `time`. Does someone have better names?
 - Add `get_basedir`
    - Does someone know a better name? This name is currently inspired from sacred.
    - Some information for the folder name convention:
      Let's say, we want that all information from one experiment is saved in the folder `.../storage/my_fance_experiment/2`.
       - The complete path is then the `storage_dir='.../storage/my_fance_experiment/2'`. This is saved as property of the trainer (`padertorch.Trainer.storage_dir`).
       - The `2` is interpreted as an experiment id. The first experiment get `1`, the second `2`, ...
       - The name `my_fance_experiment` is a name for your experiment. 
       - The enviroment variable `STORAGE_ROOT` is `'.../storage'`


boeddeker:
add Lukas style of get_new_folder with adjective_color_animal

jensheit:
Perhaps we should also rename "get_new_folder" since it is more something like "get_next_experiment_dir".

Instead of get_base_dir wouldn't it be get_experiment_dir() (especially true if we rename get_new_folder)?

boeddeker:
I think `get_new_folder` is ok. But depending on the name for `get_base_dir` a renaming may be resonable.

I don't like the combination of `get_experiment_dir` and `get_next_experiment_dir`. The first one sound like, get an existing dir, while the second get a new dir. But the first we want to use to get the parent dir.

boeddeker:
We had an offline discussion. Here our conclusion:
 - Rename `get_new_folder` to `get_new_subdir`
 - Move `get_new_subdir` to `paderbox`
 - Add cli option to `get_new_subdir`
 - Recommend a wrapper in `padertorch` instead of two functions.
   - As name we favored `get_new_storage_dir`
 - Remove `force_suffix` and `try_id` -> confusing
   - Instead introduce `prefix` and `suffix`.
 - Rename `id_type` to `id_naming`

boeddeker:
Tests fail, because https://github.com/fgnt/paderbox/pull/18 has to be merged first.

padertorch/io.py
```diff
@@ -0,0 +1,293 @@
...
+            return simu_dir
+        except FileExistsError:
+            # Catch race conditions
+            if i > 100:
+                # After some tries,
+                # expect that something other went wrong
+                raise
+
+
+def get_basedir(
```
jensheit: I do not think that we need this function here

boeddeker:
https://github.com/fgnt/paderbox/pull/18 is now merged and this PR works with paderbox master.

padertorch/io.py
```diff
@@ -0,0 +1,37 @@
...
+def get_new_storage_dir(
+        experiment_name,
+        *,
+        id_naming='index',
+        mkdir=True,
+        prefix=None,
+        suffix=None,
+        consider_mpi=False,
+        dry_run=False,
+):
```
thequilo, line 16: Ads type annotations similar to the ones in paderbox?

padertorch/io.py
```diff
@@ -0,0 +1,37 @@
...
+    """
+    This is a wrapper around `paderbox.io.new_subdir.get_new_subdir`.
+
```
thequilo, line 19: Copy the parameter docs from paderbox? I'm not sure if referencing it is enough. Especially the mpi warning is important

padertorch/io.py
boeddeker: Done

padertorch/io.py
boeddeker: Done

PR #48, opened 2020-08-18
tbx_utils: mask_to_image with color and support batch_first=None

boeddeker:
- add `color` option to `mask_to_image`
 - add option to disable `batch_first` argument (i.e. `batch_first=None`) (Better exception msg)

### DIFF
--- a/padertorch/summary/tbx_utils.py
+++ b/padertorch/summary/tbx_utils.py
@@ -113,7 +113,7 @@ def _colorize(image, color):
                 _spectrogram_to_image_cmap[color] = cmap
             except ImportError:
                 from warnings import warn
-                gray_scale = lambda x: x.transpose(1, 0)[None, ::-1, :]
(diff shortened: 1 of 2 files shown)


padertorch/summary/tbx_utils.py
```diff
@@ -68,6 +75,54 @@ def stft_to_image(signal, batch_first=False, color='viridis'):
...
+            [3, 4, 5]]])
+    """
+    if color is None:
+        return image[None, :, :]
+    else:
+        if color is True:
+            color = 'viridis'
+        try:
+            cmap = _spectrogram_to_image_cmap[color]
+        except KeyError:
```
jensheit, line 109: since we are using this exact code already at two places it should be put in a separate function

boeddeker: It is already a separate function. I removed the old code and replaced it with this function

jensheit (approved):

PR #49, opened 2020-08-19
Restructure TasNet and OR-PIT examples

thequilo:
I started a test training for both models to see if everything is still working.


padertorch/contrib/examples/source_separation/or_pit/README.md
```diff
@@ -19,13 +19,13 @@ The training procedure of the OR-PIT consists of two steps: no fine-tuning and f
 The training for the first step can be initialized with:
      
 ```bash
-$ python -m padertorch.contrib.examples.or_pit.train init with database_jsons=${path_to_your_jsons}
+$ python -m padertorch.contrib.examples.source_separation.or_pit.train init with database_jsons=${path_to_your_jsons}
  ```
```
jensheit, line 23: Shouldn't we advice the user to directly start the training without using init?

padertorch/contrib/examples/source_separation/or_pit/README.md
```diff
@@ -38,17 +38,19 @@ The fine-tune experiment can be initialized with:
...
 
 Evaluation
 ----------
 
 Start an evaluation with 
 
 ```bash
-$ python -m padertorch.contrib.examples.or_pit.evaluate init with model_path=${path_to_the_model_dir} database_json=${path_to_the_json} "datasets=['your','datasets']"
+$ python -m padertorch.contrib.examples.source_separation.or_pit.evaluate init with model_path=${path_to_the_model_dir} database_json=${path_to_the_json} "datasets=['your','datasets']"
 ```
```
jensheit, line 50: same as above, why use init instead of directly starting the evaluation?

padertorch/contrib/examples/source_separation/tasnet/README.md
```diff
@@ -12,19 +14,26 @@ With the default parameters, the following numbers can be obtained:
 
 | trained with  | SI-SDRi  | SDRi  |
 |---|---|---|
-| default  | 16.4  | 16.7  |
-| `with log_mse`  | 16.7  | 17.0  |
+| DPRNN default  | 16.4  | 16.7  |
+| DPRNN `with log_mse`  | 16.7  | 17.0  |
+
```
jensheit, line 19: we should add at least some results for the STFT/ISTFT Encoder/Decoder, since they should achieve only slightly worse results

padertorch/contrib/examples/source_separation/or_pit/README.md
thequilo: Yes, I agree, we should do that

padertorch/contrib/examples/source_separation/tasnet/README.md
thequilo: Yes. I planned to run this experiment after the test one with the TasNet encoder and decoder finishes. Or do you already have numbers produces with this model and implementation? If so, you can add them.

jensheit: I already have numbers, but not with your evaluation. To allow for a fair comparison it should be trained and evaluated with this code. 
I would add one or two sentences describing the STFT/ISTFT Enocer/Decoder and how to use them to the Readme.

jensheit: I can add those now if you don't mind

padertorch/contrib/examples/source_separation/or_pit/train.py
```diff
@@ -163,7 +173,8 @@ def prepare_iterable(
     if iterator_slice is not None:
```
boeddeker, line 173: Can you use a normal function instead of `lambda x: (x.update(num_speakers=len(x['speaker_id'])), x)[1]`?

padertorch/contrib/examples/source_separation/or_pit/train.py
```diff
@@ -163,7 +173,8 @@ def prepare_iterable(
...
         iterator = iterator[iterator_slice]
 
-    # This
+    # Group iterators by number of speakers so that all examples in a batch
+    # have the same number of speakers
     iterators = list(iterator.groupby(lambda x: x['num_speakers']).values())
 
     iterators = [
```
boeddeker, line 180:
```python
if prefetch:
    iterator = iterator.prefetch(8, 16, catch_filter_exception=True)
elif chunk_size > 0:
    iterator = iterator.catch()
```
Can you add a comment, why chunk_size >= 0 cannot raise an FilterException?

padertorch/contrib/examples/source_separation/tasnet/README.md
```diff
@@ -12,19 +14,26 @@ With the default parameters, the following numbers can be obtained:
...
 
-A storage root must be set with `export STORAGE=/path/to/your/storage`.
 After installing `padertorch`, a training can be started with
 
 ```bash
-$ python -m padertorch.contrib.examples.tasnet.train with database_json="${PATH_TO_YOUR_DATABASE_JSON}"
+$ python -m padertorch.contrib.examples.source_separation.tasnet.train with database_json="${PATH_TO_YOUR_DATABASE_JSON}"
 ```
 
+If you set the environment variable "WSO0", you don't have to provide `database_json` in the above command.
```
boeddeker, line 36: Typo: WSO0

padertorch/contrib/examples/source_separation/or_pit/README.md
thequilo: Done

padertorch/contrib/examples/source_separation/or_pit/train.py
thequilo: Done

padertorch/contrib/examples/source_separation/or_pit/train.py
thequilo: Done

jensheit:
> Can you change `or` to `and/or`? Your example is `and`.

it was supposed to be a mathematical 'or' but sure I can change that

thequilo:
I made some changes:
 - Use `InputMetrics` and `OutputMetrics` from `pb_bss`
 - I added `ompi -V gcc/2.0.2-mt` which requires to load the correct MPI version, but other versions fail sometimes
 - Moved code to compute the means over the results to a separate function

jensheit:
Perhaps you could rename the tasnet.py to model.py so that we can keep the naming similar in other examples as well

padertorch/contrib/examples/source_separation/or_pit/README.md
```diff
@@ -38,17 +38,19 @@ The fine-tune experiment can be initialized with:
 $ make fine-tune
 ```
 
-It uses the same configuration (including number of epochs, data, etc.) as the base training.
+It creates a new storage dir and uses the same configuration (including number of epochs, data, etc.) as the base training.
```
TCord, line 41: It's a bit unclear what "It" means in this context.

PR #51, opened 2020-08-26
Add data preparation / JSON generation script for WSJ0-mix

thequilo:


padertorch/contrib/data/wsj0_mix/prepare_data.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
+# Put your paths here!
+
+# Path to the database as generated by the matlab scripts
+database_path=
+
+# If you need transcriptions, put the path to the WSJ root folder here
+wsj0_root=
```
boeddeker, line 9: Can you adopt the idea from https://github.com/espnet/espnet/blob/master/egs2/TEMPLATE/asr1/db.sh ?
In this way, we can specify some defaults for us.

padertorch/contrib/data/wsj0_mix/prepare_data.sh
```diff
@@ -0,0 +1,17 @@
...
+
+python -m padertorch.contrib.data.wsj0_mix.create_json \
+	--database_path ${database_path} \
+	--wsj0_root ${wsj0_root} \
+	--json_path wsj0_mix_8k.json \
```
boeddeker, line 14: Why do you exclude `min` in the json name?

boeddeker:
I just scrolled over the code. I am not sure, what you changed compared to our internal code.

padertorch/contrib/data/wsj0_mix/prepare_data.sh
thequilo: Yes, what is the best hostname to check for? Is `[[ "$(hostname -d)" == "nt.uni-paderborn.de" ]];` enough, or shall we include the PC2 cluster as well?

thequilo:
> I just scrolled over the code. I am not sure, what you changed compared to our internal code.

I changed many things. Here is a short list:
- Remove code to determine the shortest speaker
- Iterate over directory structure instead of mix file
- Make spk2gender optional
- Download `normalize_transcript.pl` from Kaldi for normalizing the transcripts
- Use `pb.utios.process_caller.run_process` for direct piping of stdin and stdout to `normalize_transcript.pl`, instead of temporary files
- Everything that is not optional is now created in one big dictionary which makes the code a lot shorter
- Remove `global` variables
- Rearrange and rename some input arguments so that they match the above modifications

padertorch/contrib/data/wsj0_mix/prepare_data.sh
thequilo: For no reason. I can include it

padertorch/contrib/data/wsj0_mix/create_json.py
```diff
@@ -0,0 +1,273 @@
...
+OBSERVATION = 'observation'
+SPEECH_SOURCE = 'speech_source'
+AUDIO_PATH = 'audio_path'
+
+
+def download_normalize_transcript():
+    """
+    Downloads Kaldis "normalize_transcript.pl" for WSJ to the working directory
+    """
+    normalize_transcript_path = Path('normalize_transcript.pl').resolve()
```
boeddeker, line 52: Can you use 
```
from appdirs import user_cache_dir
user_cache_dir('paderbox')
```
instead of `cwd`?

boeddeker: Or alternatively, write the file next to the json.

boeddeker:
> Iterate over directory structure instead of mix file

Can you add some assert to the code to identify broken folders?
e.g. the number of files and the number of examples per dataset.
In `sms_wsj` we observed that there can be many differences and many buggy folders.

padertorch/contrib/data/wsj0_mix/prepare_data.sh
boeddeker: Yes, that is fine for the workstations. For the clusters I would suggest to check `$PC2SYSNAME` if it is `OCULUS`.

boeddeker:
> Changed to new version of database (without _2 postfix in filenames)

Are there more changes between the wsj0-2mix version?
If not, how about to support both versions? As far as I know, is the only relevant change that they removed the `_2`.

thequilo:
> > Changed to new version of database (without _2 postfix in filenames)
> 
> Are there more changes between the wsj0-2mix version?
> If not, how about to support both versions? As far as I know, is the only relevant change that they removed the `_2`.

At least that's the only obvious change that I am aware of. There might be slight differences in the data depending on the version of voicebox used to create the older version. I can try to support both of them

padertorch/contrib/data/wsj0_mix/prepare_data.sh
thequilo: Done

padertorch/contrib/data/wsj0_mix/create_json.py
thequilo: Done

padertorch/contrib/data/wsj0_mix/prepare_data.sh
```diff
@@ -1,12 +1,20 @@
 #!/usr/bin/env bash
 
 # Put your paths here!
+if [[ "$(hostname -d)" == nt.uni-paderborn.de ]]; then
+	# NT specific defaults
+	database_path=/net/db/merl_speaker_mixtures/data
+	wsj0_root=/net/db/wsj
+elif [[ "${PC2SYSNAME}" == OCULUS ]]; then
+	database_path=/scratch/hpc-prf-nt1/cbj/net/db/merl_speaker_mixtures
```
boeddeker, line 9: Is the `data` subfolder missing?

boeddeker (approved):

PR #56, opened 2020-09-09
update mask estimator example to use NT_DATABASE_JSON_DIR variable

jensheit:


padertorch/contrib/examples/speech_enhancement/mask_estimator/evaluate.py
```diff
@@ -45,20 +51,12 @@ def config():
...
         raise EnvironmentError(
-            'You have to specify an STORAGE_ROOT '
-            'environmental variable see getting_started'
-        )
-    elif not Path(storage_root).exists():
-        raise FileNotFoundError(
-            'You have to specify an existing STORAGE_ROOT '
-            'environmental variable see getting_started.\n'
-            f'Got: {storage_root}'
+            'You have to specify an STORAGE_ROOT environmental variable'
```
thequilo, line 56: By the way, "an" is wrong here, it should be "a" ;)
Or the word "existing" might be missing

Issue #43, opened 2020-08-11
restructure and update the contrib examples


jensheit:
Split example directory into toy examples and task specific examples.
Update the example training scripts so that the experiments use similar directory structures.
Add evaluation scripts if missing and update the evaluation directory structure, so that the evaluation files are written in a new directory inside the training directory.

- [x] create io.py and add function for default training directory definition @boeddeker 
- [x] move to toy examples and update mnist, multi_gpu, configurable @boeddeker 
- [x] remove the acoustic model example . @boeddeker 
- [x] rename and update the simple train script for audio tagging, remove the advanced training script @JanekEbb 
- [x] add evaluation script for audio tagging @JanekEbb 
- [x] move, rename and update the training script for mask estimator @jensheit
- [x] add evaluation script for mask estimator @jensheit 
- [x] move, rename and update the or_pit and tasnet example @thequilo @jensheit 
- [x] move and update the pit example @TCord 
- [x] update the speaker classification example @michael-kuhlmann  @JanekEbb  
- [x] add evaluation script for speaker classification @michael-kuhlmann 
- [x] move and update the wavenet example @JanekEbb

PR #57, opened 2020-09-10
Update speaker classification example

michael-kuhlmann:
Contribution to issue #43


padertorch/contrib/examples/speaker_classification/supervised/README.md
```diff
@@ -0,0 +1,63 @@
...
+                ...
+            },
+            ...
+        },
+        ...
+    }
+}
+```
+If you train on a different database than LibriSpeech, you can replace
+`<chapter-id>` with an arbitrary string.
```
jensheit, line 43: what is chaper_id here?

padertorch/contrib/examples/speaker_classification/supervised/README.md
```diff
@@ -0,0 +1,63 @@
...
+It requires [dlp_mpi](https://github.com/fgnt/dlp_mpi) to be installed.
+For each misclassified example, symlinks to the example audio file and to an audio
+example of the wrongly classified speaker are stored.
+
+## Results
+
+| Database | Dataset | Num. Speakers | Num. Eval Examples | Classification Accuracy |
+| :------: | :-----: | :-----------: | :----------------: | :---------------------: |
+| LibriSpeech | clean_100 | 251 | 2853 | 98.60% |
+| LibriSpeech | clean_360 | 921 | 10401 | 94.72% |
```
jensheit, line 63: I am not sure what to take from these numbers, have you trained on clean_460 and this are the results on part of the training set? Why are there no results on dev or eval?

padertorch/contrib/examples/speaker_classification/supervised/data.py
```diff
@@ -0,0 +1,101 @@
+import numpy as np
+
+from lazy_dataset.database import JsonDatabase
+
+from padertorch.contrib.je.data.transforms import (
+    LabelEncoder, AudioReader, STFT, MelTransform, Collate
```
jensheit, line 6: is this STFT something different compared to the STFT in paderbox.transform.STFT?

padertorch/contrib/examples/speaker_classification/supervised/data.py
```diff
@@ -0,0 +1,101 @@
...
+    indices = np.arange(num_examples)
+
+    dev_size = int(num_examples * dev_split)
+    test_size = int(num_examples * test_split)
+
+    test_candidates = r.choice(indices, size=test_size, replace=False)
+    indices = np.delete(indices, test_candidates)
+    dev_candidates = r.choice(indices, size=dev_size, replace=False)
+    train_candidates = np.delete(indices, dev_candidates)
+
```
jensheit, line 55: why are you creating a test and dev set from train and do not use the original dev and test set?

padertorch/contrib/examples/speaker_classification/supervised/train.py
```diff
@@ -0,0 +1,119 @@
...
+        norm='batch'
+    )
+    gru = GRU(input_size=512, hidden_size=256, num_layers=2, batch_first=True)
+    fcn = fully_connected_stack(
+        256, hidden_size=[256], output_size=num_speakers, dropout=0.
+    )
+
+    speaker_clf = SpeakerClf(feature_extractor, cnn, gru, fcn)
+    return speaker_clf
+
```
jensheit, line 66: Why not move this into the config? Since that would be more similar to our real workflow

padertorch/contrib/examples/speaker_classification/supervised/README.md
michael-kuhlmann: I have trained separately on the clean_100 and clean_360 datasets. The network learns to map an input example to a finite number of output labels (here, 251 or 921). Therefore, it cannot handle speakers not seen during training without fine-tuning on new speakers. As the datasets of LibriSpeech all contain different speakers, evaluation was performed on a subset of the dataset used in training, i.e., I have split the clean_100 dataset into three parts: 80% are used for training, 10% for validation and 10% are held out for evaluation on which the numbers in the README are reported. The utterances of the evaluation set were thus not seen during training.
I have no comparisons to state-of-the-art results. I could look out for them, but for an example, I guess the achieved classification accuracies are good enough? Maybe @JanekEbb has some comments on this.

padertorch/contrib/examples/speaker_classification/supervised/train.py
michael-kuhlmann: I guess I could do that. I kept this from the previous version were sacred was not yet used and missed to move that part. I'll change that.

padertorch/contrib/examples/speaker_classification/supervised/data.py
michael-kuhlmann: It inherits from paderborn.transform.STFT and modifies the __call__ function to handle dictionary inputs in addition to numpy arrays, i.e., you can directly pass an instance of this class to the `map()` call of lazy_dataset.

jensheit: The question it entails is how much work would it be to write some kind of prepare function which is less dependent on contrib and just uses paderbox and the main padertorch. If this would be a lot of work it is not necessary, I just think it is nice to have less dependency on contrib functions which can change all the time.

padertorch/contrib/examples/speaker_classification/supervised/train.py
michael-kuhlmann: Updated in commit https://github.com/fgnt/padertorch/pull/57/commits/697cb3eb65d506f4df19ee730135936b52cd5f92

padertorch/contrib/examples/speaker_classification/supervised/data.py
michael-kuhlmann: Updated with commit https://github.com/fgnt/padertorch/pull/57/commits/6988846537bc623d6b26664450e7d7f243454e2f. Only `LabelEncoder` remains dependent of `contrib` since there is no equivalent in the core code.

padertorch/contrib/examples/speaker_classification/supervised/train.py
jensheit: Here class_to_str is not necessary, if you call SpeakerClf.get_config(model). 
This should be done because otherwise all values not specified in this config are not written to the config.json. This might be a problem in case of changing defaults.

You could think about moving the options from the trainer to the config as well.

padertorch/contrib/examples/speaker_classification/supervised/data.py
```diff
@@ -0,0 +1,118 @@
...
+    ds = ds.map(prepare_example)
+
+    speaker_encoder = LabelEncoder(
+        label_key='speaker_id', storage_dir=storage_dir, to_array=True
+    )
+    speaker_encoder.initialize_labels(dataset=ds, verbose=True)
+    ds = ds.map(speaker_encoder)
+
+    train_set, validate_set, test_set = train_test_split(ds)
```
boeddeker, line 31: Could you add a small comment to the code, why you split the train set?

padertorch/contrib/examples/speaker_classification/supervised/data.py
```diff
@@ -0,0 +1,118 @@
...
+    example = pt.data.utils.collate_fn(example)
+    example['features'] = pad_sequence(
+        example['features'], batch_first=True
+    ).transpose(-1, -2)
+    example['speaker_id'] = np.stack(example['speaker_id'])
+    return example
+
+
+def prepare_dataset(dataset, batch_size=16, training=False, prefetch=True):
+    dataset = dataset.map(_prepare_features)
```
boeddeker, line 89: Is it not easier to read the code, when you do only one mapping to the dataset?

padertorch/contrib/examples/speaker_classification/supervised/data.py
```diff
@@ -0,0 +1,118 @@
...
+            _example['audio_path'] = example['audio_path']
+        return _example
+
+    dataset = dataset.map(finalize)
+
+    if training:
+        dataset = dataset.shuffle(reshuffle=True)
+    if prefetch:
+        dataset = dataset.prefetch(
+            num_workers=8, buffer_size=10*batch_size
```
boeddeker, line 110: 160 is a huge bugger size. Is the data so unbalanced, that `batch_dynamic_time_series_bucket` needs such a large buffer size?

padertorch/contrib/examples/speaker_classification/supervised/evaluate.py
```diff
@@ -0,0 +1,166 @@
...
+def defaults():
+    model_path = None
+    assert model_path is not None, (
+        'model_path cannot be None.\n'
+        'Start the evaluation with "python -m padertorch.contrib.examples.'
+        'speaker_classification.evaluate with model_path=<path/to/trained/model>"'
+    )
+    load_ckpt = 'ckpt_best_loss.pth'
+    batch_size = 1
+    device = 'cpu'
```
boeddeker, line 43: You could change this to `0 if torch.cuda.is_available() else 'cpu'`. Then it automatically detects the gpu. (i.e. it reads CUDA_VISIBLE_DEVICES)

padertorch/contrib/examples/speaker_classification/supervised/evaluate.py
```diff
@@ -0,0 +1,166 @@
...
+        prefetch=device != 'cpu'
+    )
+    with torch.no_grad():
+        summary = dict(
+            misclassified_examples=dict(),
+            correct_classified_examples=dict(),
+            hits=list()
+        )
+        for batch in split_managed(
+            test_set, is_indexable=batch_size == 1 and device == 'cpu',
```
boeddeker, line 80: Be careful with `is_indexable=batch_size == 1`. If `is_indexable` is `False`, and you use MPI, then each process iterates over the dataset. This is expensive, because you do all the data loading in the dataset.

padertorch/contrib/examples/speaker_classification/supervised/train.py
boeddeker: You could call `pt.Configurable.get_config(model)` after you defined the model, that would fill the config with all defaults.
i.e. the config would then be more verbose.

I think he calls `class_to_str`, because he uses sacred to save the config.
You could use `pt.configurable.recursive_class_to_str` before you write the config, then the values would also be str, when you write the config.

michael-kuhlmann: Should now be considered by commit https://github.com/fgnt/padertorch/pull/57/commits/7f9eaaf7c740f13092c3fad36cfb457c93bbeafb

padertorch/contrib/examples/speaker_classification/supervised/data.py
michael-kuhlmann: https://github.com/fgnt/padertorch/pull/57/commits/a3688c1ebb4ac61cf734e2428cd578027456742e

padertorch/contrib/examples/speaker_classification/supervised/data.py
michael-kuhlmann: https://github.com/fgnt/padertorch/pull/57/commits/3654cbd02a18b9be271f20ea1898ccd5c4cc3f81

padertorch/contrib/examples/speaker_classification/supervised/evaluate.py
michael-kuhlmann: Thanks for the hint. Now, the dataset is always indexable if evaluation is performed on the CPU (https://github.com/fgnt/padertorch/pull/57/commits/5517ed73f58728726506cfafd225333a8aed08c7).

padertorch/contrib/examples/speaker_classification/supervised/evaluate.py
michael-kuhlmann: https://github.com/fgnt/padertorch/pull/57/commits/22310de5eafca150b8fcd944034768e0abcf8d98

padertorch/contrib/examples/speaker_classification/supervised/data.py
michael-kuhlmann: I took the value from @JanekEbb and did not experiment with other values. However, the dataset looks unbalanced to me (plot from padercontrib.database.get_bucket_boundaries) :
Bildschirmfoto 2020-09-21 um 12 43 26

boeddeker: Ah, ok. I just wanted to ensure, it is intended. Depending on the size of the examples, you could run out of memory.

jensheit (approved):

PR #59, opened 2020-09-11
Add command "test_run" to all source_separation examples

thequilo:
@jensheit is this ok for your example tests?


jensheit (approved):
This should work for the test case in combination with your last update

PR #63, opened 2020-09-18
Configurable improvements

boeddeker:
- `import_class`: Support nested factories (e.g. methods of class)
 - `recursive_class_to_str`: Add sort option (i.e. use the order of the signature)
 - `config_to_instance`: Improve exception msg and add strict option
 - Add `padertorch.io.dump_config` and `padertorch.io.load_config`


padertorch/configurable.py
```diff
@@ -523,18 +523,26 @@ def fix_doctext_import_class(locals_dict):
...
+        '...configurable.foo'
+        >>> Foo  # doctest: +ELLIPSIS
+        <class '...configurable.Foo'>
         >>> import_class(class_to_str(foo)) == foo
         True
-        >>> import_class(class_to_str(Foo))
-        <class 'padertorch.configurable.Foo'>
+        >>> import_class(class_to_str(Foo))  # doctest: +ELLIPSIS
+        <class '...configurable.Foo'>
     """
```
jensheit, line 546: why do you need elipsis here? Shouldnt Configurable always be imported from padertorch?

padertorch/io.py
```diff
@@ -79,3 +80,103 @@ def get_new_storage_dir(
...
+        path
+):
+    import paderbox as pb
+    return pb.io.load(path)
+
+
+def loads_config(
+        content,
+        format=None,  # '.yaml' or '.json'
+):
```
jensheit, line 158: this should be the same default as above in dumps_config

padertorch/io.py
```diff
@@ -79,3 +80,103 @@ def get_new_storage_dir(
...
+            format = '.yaml'
+
+    if format == ".json":
+        from paderbox.io import loads_json
+        return loads_json(content)
+    elif format == ".yaml":
+        from paderbox.io import loads_yaml
+        return loads_yaml(content)
+    else:
+        raise NotImplementedError(format)
```
jensheit, line 182: missing new line

padertorch/configurable.py
boeddeker: This is a bug in pycharm. In pycharm it is `configurable.Foo`, while pytest has the correct `padertorch.configurable.Foo`.

The bug is reported, but they don't care:
https://youtrack.jetbrains.com/issue/PY-26738

padertorch/io.py
boeddeker: The default is compatible to `dumps_config`, so I thought this is fine.

jensheit: but why take None here and '.yaml' for dumps_config?

padertorch/io.py
boeddeker: done

padertorch/io.py
boeddeker: Since we have a limited number of supported serialization formats and the outer type is a dict, we can simply guess the format (i.e. json starts with `{`, while yaml doesn't.)

jensheit: Shouldn't the default for ```loads_config``` still be None and load_config does the guessing?

jensheit: But sure I see your point. It just looks a little bit strange to have different defaults. But its fine

jensheit (approved):

padertorch/io.py
boeddeker: In `load_config` it is not necessary to guess, because the suffix tells which format it is.
I just wrote the doctests and it was annoying to specify the format, so I added the guessing to `loads_config`.

If it is to much magic, I can remove it (I expect, not end user will use `loads` and `dumps`).

boeddeker: I removed not the auto-detection

PR #69, opened 2020-09-28
switches to if statement to reduce memory requirement for backwards step

jensheit:
Adjusts changes in #68 to account for @boeddeker request to reduce memory in backwards step


boeddeker (approved):
Thanks

PR #70, opened 2020-10-15
adds an output activation to fully_connected_stack

jensheit:


padertorch/modules/fully_connected.py
```diff
@@ -11,7 +11,8 @@ def fully_connected_stack(
         hidden_size: List[int],
         output_size: int,
         activation: str = 'relu',
-        dropout: float = 0.5
+        dropout: float = 0.5,
+        output_activation: str = None
```
thequilo, line 15: Add comma at end of line?

thequilo (approved):

padertorch/modules/fully_connected.py
jensheit: good catch, I will add it

boeddeker (approved):

jensheit:
We could discuss whether we should raise an error if the stack has the length of one and the activation is not set to None, because in this case the activation is ignored.

PR #71, opened 2020-10-20
adds CallableDispatcher 

jensheit:
Adds a CallableDispatcher to mapping and removes mappings which are not used anywhere at the moment.

### DIFF
--- a/padertorch/ops/mappings.py
+++ b/padertorch/ops/mappings.py
@@ -9,6 +9,31 @@ __all__ = [
     'DTYPE_MAP',
 ]
 
+class CallableDispatcher(Dispatcher):
+    """
+       Is basically a dict for non-callable inputs
+       with a better error message on key error.
+       If the input is a callable it is returned.
+       >>> from padertorch.ops.mappings import CallableDispatcher
+       >>> d = CallableDispatcher(abc=1, bcd=2)
+       >>> d['acd']  #doctest: +ELLIPSIS
+       Traceback (most recent call last):
+       ...
+       paderbox.utils.mapping.DispatchError: Invalid option 'acd'.
+       Close matches: ['bcd', 'abc'].
+       >>> from padertorch.ops.mappings import CallableDispatcher
+       >>> d = CallableDispatcher(abc=1, bcd=2)
+       >>> d[np.median]  #doctest: +ELLIPSIS
+       <function median at ...
+       """
+
+    def __getitem__(self, item):
+        if callable(item):
+            return item
+        else:
+            return super().__getitem__(item)
+
+
 ACTIVATION_FN_MAP = Dispatcher(
     relu=torch.nn.ReLU,
     prelu=torch.nn.PReLU,
--- a/padertorch/ops/mappings.py
+++ b/padertorch/ops/mappings.py
@@ -43,6 +43,9 @@ ACTIVATION_FN_MAP = Dispatcher(
     identity=torch.nn.Sequential,  # https://github.com/pytorch/pytorch/issues/9160
 )
 
+
+# These mappings are not used at the moment if required they can be added again
+# but the naming convention shuld be updated.
 # POOLING_FN_MAP = CallableDispatcher(
 #     median=np.median,
 #     average=np.mean,
--- a/padertorch/ops/mappings.py
+++ b/padertorch/ops/mappings.py
@@ -32,7 +32,7 @@ class CallableDispatcher(Dispatcher):
             return super().__getitem__(item)
 
 
-ACTIVATION_FN_MAP = Dispatcher(
+ACTIVATION_FN_MAP = CallableDispatcher(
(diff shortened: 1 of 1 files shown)


padertorch/ops/mappings.py
```diff
@@ -5,11 +5,34 @@
 
 __all__ = [
     'ACTIVATION_FN_MAP',
-    'POOLING_FN_MAP',
-    'DTYPE_MAP',
 ]
 
-ACTIVATION_FN_MAP = Dispatcher(
+class CallableDispatcher(Dispatcher):
```
thequilo, line 10: Would it be better to move this to `paderbox.utils.mapping`, next to `Dispatcher`?

padertorch/ops/mappings.py
```diff
@@ -5,11 +5,34 @@
...
+    """
+       Is basically a dict for non-callable inputs
+       with a better error message on key error.
+       If the input is a callable it is returned.
+       >>> from padertorch.ops.mappings import CallableDispatcher
+       >>> d = CallableDispatcher(abc=1, bcd=2)
```
boeddeker, line 16: Should one example work?

padertorch/ops/mappings.py
boeddeker, line 10: Make it private? For public I don't like the name.

padertorch/ops/mappings.py
```diff
@@ -20,29 +43,32 @@
...
-)
-
-OPTIMIZER_MAP = Dispatcher(
-    sgd=optim.SGD,
-    adam=optim.Adam,
-    adagrad=optim.Adagrad
-)
+# These mappings are not used at the moment if required they can be added again
+# but the naming convention shuld be updated.
+# POOLING_FN_MAP = CallableDispatcher(
```
boeddeker, line 49: change to `REDUCE_MAP`

boeddeker: Sorry, `NP_REDUCE_MAP`

padertorch/ops/mappings.py
```diff
@@ -20,29 +43,32 @@
...
+#     median=np.median,
+#     average=np.mean,
+#     min=np.min,
+#     max=np.max,
+# )
+#
+# TORCH_POOLING_FN_MAP = CallableDispatcher(
```
boeddeker, line 56: Change to `REDUCE_MAP`

padertorch/ops/mappings.py
jensheit: this is an example taken from the original dispatcher since it behaves like a dictionary if it works, which is not interesting

padertorch/ops/mappings.py
```diff
@@ -5,11 +5,34 @@
...
 __all__ = [
     'ACTIVATION_FN_MAP',
-    'POOLING_FN_MAP',
-    'DTYPE_MAP',
 ]
 
-ACTIVATION_FN_MAP = Dispatcher(
+class CallableDispatcher(Dispatcher):
+    """
+       Is basically a dict for non-callable inputs
```
thequilo, line 12: This formulation is confusing. First time I read it I thought "Why is the CallableDispatcher not used for callable inputs?". I think the other way round, like "For non-callable inputs, ... . For callable inputs, ...", is less confusing

padertorch/ops/mappings.py
jensheit: I have not updated the names since the comments states that the names are not up to date

padertorch/ops/mappings.py
jensheit: I can do that, however @boeddeker wants to make it privat which would not realy make sense in that case.

padertorch/ops/mappings.py
jensheit: done

padertorch/ops/mappings.py
jensheit: done

padertorch/ops/mappings.py
jensheit: done

padertorch/ops/mappings.py
thequilo: Yes, I agree that making it private and at the same time moving it doesn't make sense. We have to discuss if there are other use-cases outside of `pt.ops.mappings`

padertorch/ops/mappings.py
boeddeker: Ups, I didn't read the comment.

padertorch/ops/mappings.py
boeddeker: And we also need a good name for it.
I don't like the current one. But for a private one, it would be ok.

padertorch/ops/mappings.py
boeddeker: ok

PR #72, opened 2020-10-27
Adds ConvNet to the TasNet example

jensheit:
Adds ConvNet to padertorch.modules
Rewrite tasnet.train to force user to specify separator
Add ConvNet to tasnet.README.md

Open steps:
Use JE Conv implementation
Replace contrib.jensheit.norm with calls to modules.normalization


padertorch/contrib/examples/source_separation/tasnet/train.py
```diff
@@ -142,6 +134,29 @@ def stft():
...
+        'window_length': 100,
+        'hop_size': 50,
+        'num_blocks': 6,
+    }}}
+
+
+@ex.named_config
+def convnet():
+    feat_size = 256
+    trainer = {'model': {'separator': dict(
```
thequilo, line 152: Why is this `dict`, but for the dprnn it is `{}`?

jensheit: it's due to copy and paste, but I change it to be more similar

PR #73, opened 2020-11-02
fix regression functions: They ignored the reduce agrument

boeddeker:
- `reduce='none'` never worked in Pytorch, because `reduce` is interpreted as `bool` and `bool('none') is False`.
 - rename `reduce` to `reduction` to follow Pytorch names.
   - The breaking change is intended here (I wanted to break all code, but that does not work)

### DIFF
--- a/padertorch/ops/losses/regression.py
+++ b/padertorch/ops/losses/regression.py
@@ -3,10 +3,10 @@ import torch
 from torch.nn import functional as F
 
 
-def _get_scaling_factor(targets, estimates):
+def _get_scaling_factor(target, estimate):
     return torch.unsqueeze(torch.einsum(
-        '...t,...t->...', estimates, targets
-    ), -1) / torch.norm(targets, dim=-1, keepdim=True) ** 2
+        '...t,...t->...', estimate, target
+    ), -1) / torch.norm(target, dim=-1, keepdim=True) ** 2
 
 
 def _reduce(array, reduction):
@@ -150,9 +150,9 @@ def si_sdr_loss(estimate, target, reduction='mean', offset_invariant=False,
     >>> np.random.seed(0)
     >>> reference = torch.tensor(np.random.randn(100))
 
-    >>> def calculate(estimates, targets):
-    ...     print('Torch loss:', si_sdr_loss(estimates, targets))
-    ...     print('Numpy metric:', si_sdr(estimates.numpy(), targets.numpy()))
+    >>> def calculate(estimate, target):
+    ...     print('Torch loss:', si_sdr_loss(estimate, target))
+    ...     print('Numpy metric:', si_sdr(estimate.numpy(), target.numpy()))
 
     Perfect estimation
     >>> calculate(reference, reference)
--- a/padertorch/ops/losses/regression.py
+++ b/padertorch/ops/losses/regression.py
@@ -271,14 +271,3 @@ def log1p_mse_loss(estimate: torch.Tensor, target: torch.Tensor,
 
 # TODO: Add log1p_mse loss from interspeech paper
 # TODO: remove _reduce
-
-
-time_domain_loss_functions = {
-    'log-mse': log_mse_loss,
-    'si-sdr': si_sdr_loss,
-    'si-sdr-grad-stop': partial(si_sdr_loss, grad_stop=True),
-    'mse': F.mse_loss,
-    'sdr': sdr_loss,
-}
-
-
--- a/padertorch/ops/losses/regression.py
+++ b/padertorch/ops/losses/regression.py
@@ -21,6 +21,33 @@ def _reduce(array, reduction):
             f'Unknown reduction: {reduction}. Choose from "sum", "mean".')
 
 
+def mse_loss(estimate: torch.Tensor, target: torch.Tensor,
+             reduction: str = 'sum'):
+    """
+    Computes the mse loss.
+    The `reduction` only affects the speaker dimension; the time dimension is always
+    reduced by a mean operation as in [1].
+
+    Args:
+        estimate (... x T): The estimated signal
+        target (... x T, same as estimate): The target signal
+        reduction: 'mean', 'sum' or 'none'/None for batch dimensions
+
+    Returns:
+
+    >>> estimate = [[1., 2, 3], [4, 5, 6]]
+    >>> target = [[2., 3, 4], [4, 0, 6]]
+    >>> mse_loss(torch.tensor(estimate), torch.tensor(target))
+    tensor(0.9208)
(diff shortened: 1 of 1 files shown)


padertorch/ops/losses/regression.py
```diff
@@ -9,20 +9,20 @@ def _get_scaling_factor(targets, estimates):
...
         return array
-    if reduce == 'sum':
+    if reduction == 'sum':
         return torch.sum(array)
-    elif reduce == 'mean':
+    elif reduction == 'mean':
         return torch.mean(array)
     else:
-        raise ValueError(f'Unknown reduce: {reduce}. Choose from "sum", '
+        raise ValueError(f'Unknown reduce: {reduction}. Choose from "sum", '
```
thequilo, line 20: Change to "Unknown reduction"

padertorch/ops/losses/regression.py
```diff
@@ -49,16 +49,23 @@ def log_mse_loss(estimate: torch.Tensor, target: torch.Tensor,
...
-        F.mse_loss(estimate, target, reduce='none').mean(dim=-1).log10(),
-        reduce=reduce
+        F.mse_loss(estimate, target, reduction='none').mean(dim=-1).log10(),
+        reduction=reduction
     )
 
 
 def sdr_loss(estimates: torch.Tensor, targets: torch.Tensor,
-             reduce: str = 'mean'):
+             reduction: str = 'mean', axis=-1):
```
thequilo, line 68: Call this `dim` to also follow PyTorch?

padertorch/ops/losses/regression.py
boeddeker: Done

padertorch/ops/losses/regression.py
boeddeker: Good point. Most pytorch functions support both.
But this change was not intended. I removed the axis argument.

padertorch/ops/losses/regression.py
```diff
@@ -49,45 +49,59 @@ def log_mse_loss(estimate: torch.Tensor, target: torch.Tensor,
...
+    tensor(-6.5167)
+    >>> sdr_loss(torch.tensor(estimate), torch.tensor(target), reduction=None)
+    tensor([-9.8528, -3.1806])
+
     """
     # Calculate the SNR. The square in the power computation is moved to the
     # front, thus the 20 in front of the log
     snr = 20 * torch.log10(
-        torch.norm(targets, dim=-1) / torch.norm(estimates - targets, dim=-1)
+        torch.norm(targets, dim=axis) / torch.norm(estimates - targets, dim=axis)
```
thequilo, line 90: Looks like `axis` is not defined here

thequilo: And `estimates` and `targets` are not defined

padertorch/ops/losses/regression.py
```diff
@@ -49,45 +49,59 @@ def log_mse_loss(estimate: torch.Tensor, target: torch.Tensor,
...
-        reduce=reduce
+        F.mse_loss(estimate, target, reduction='none').mean(dim=-1).log10(),
+        reduction=reduction
     )
 
 
-def sdr_loss(estimates: torch.Tensor, targets: torch.Tensor,
-             reduce: str = 'mean'):
+def sdr_loss(estimate: torch.Tensor, target: torch.Tensor,
+             reduction: str = 'mean'):
```
thequilo, line 68: How about naming them `input` and `target`, similar to many PyTorch loss functions? Also for the other loss functions here

padertorch/ops/losses/regression.py
boeddeker: Done

padertorch/ops/losses/regression.py
boeddeker: I am not sure, if `input` is a good name. The name hides a built-in.

boeddeker:
I added `mse_loss` and removed `time_domain_loss_functions`

padertorch/ops/losses/regression.py
thequilo: I agree, I don't like the name either. That's the reason why it's not called input already. But I thought it might be good to use the same names as PyTorch.

padertorch/ops/losses/regression.py
```diff
@@ -3,26 +3,53 @@
...
+
+    >>> estimate = [[1., 2, 3], [4, 5, 6]]
+    >>> target = [[2., 3, 4], [4, 0, 6]]
+    >>> mse_loss(torch.tensor(estimate), torch.tensor(target))
+    tensor(0.9208)
+    >>> mse_loss(torch.tensor(estimate), torch.tensor(target), reduction=None)
+    tensor([0.0000, 0.9208])
+    """
+    return _reduce(
+        F.mse_loss(estimate, target, reduction='none').mean(dim=-1).log10(),
```
thequilo, line 46: `log10` shouldn't be there for `mse_loss`

padertorch/ops/losses/regression.py
```diff
@@ -49,45 +76,59 @@ def log_mse_loss(estimate: torch.Tensor, target: torch.Tensor,
...
+    >>> log_mse_loss(torch.tensor(estimate), torch.tensor(target))
+    tensor(0.9208)
+    >>> log_mse_loss(torch.tensor(estimate), torch.tensor(target), reduction=None)
+    tensor([0.0000, 0.9208])
     """
     # Use the PyTorch implementation for MSE, should be the fastest
     return _reduce(
-        F.mse_loss(estimate, target, reduce='none').mean(dim=-1).log10(),
-        reduce=reduce
+        F.mse_loss(estimate, target, reduction='none').mean(dim=-1).log10(),
```
thequilo, line 89: I think the mean should be a sum to be the same as the Demystifying TasNet paper

padertorch/ops/losses/regression.py
boeddeker: Yes, fixed

padertorch/ops/losses/regression.py
boeddeker: I'd like to keep the mean, here, otherwise it is not a mean square error.
It has still the same gradient as the loss denoted in the Demystifying TasNet paper.
So this change has no large impact. (Moving the 1/T outside the log it is only a bias)

I don't know how the mean operation is implemented,
but in theory the mean operation can be more stable (e.g. avoid overflow).
(Probably it relies on sum and is not more stable)

padertorch/ops/losses/regression.py
jensheit: I think it is more important to have the same order than the same name, so I would keep the names as they are.

thequilo (approved):

Issue #61, opened 2020-09-15
Configurable: Support positional only arguments


boeddeker:
At the moment is any callable a configurable, when it supports key value arguments.
While this is verbose and most factories and classes are supported, there are a few we don't support.

Maybe the most relevant example for `padertorch` is `torch.nn.Sequential(*args)`.
To use this class, at the moment we have to use a wrapper around that class.
A native support would be nice.

In a small group, we discussed this offline, but haven't found the solution, that we want to realize.

First priority is, that the implementation is no breaking change, so all examples that follow, won't break current configs.

Here some examples, how it could be realized:

First, lets assume, we have these factories:
```python
def foo(*numbers):
    ...
def bar(a, b, \):  # positional only, like operator.add
    ...
```
and we want to call them as
```python
foo(1, 2)
bar(1, 2)
```

now follow some ideas, how the config could look like:

# 1. Reserved keyword 'args', 'factory_args' or '*'
```
{'factory': 'foo', 'args': [1, 2]}
{'factory': 'bar', 'args': [1, 2]}
```
```
{'factory': 'foo', '*': [1, 2]}
{'factory': 'bar', '*': [1, 2]}
```

# 2. Signature check with "assignment"
Ignore that the arguments is a positional only in the config and simply do an assignment style in the config.
In the implementation, we then to the mapping to the positional argument.
```
{'factory': 'foo', 'numbers': [1, 2]}
{'factory': 'bar', a=1, b=2]}
```
```
{'factory': 'foo', '*numbers': [1, 2]}
{'factory': 'bar', a=1, b=2]}
```

# 3. Lisp style
The factory can be a list, the first argument is then the function, while the others are the arguments:
```
{'factory': ['foo', 1, 2]}
{'factory': ['foo', 1, 2]}
```

# My opinion

The 2. is nice for positional only arguments like `operator.add`.
But this rarely happen in practice, because it was only supported for C and C++ functions until py37 (PEP 570).
For `*args` I don't like it. It looks strange and is not robust against renaming.

For the first my favorite key would be `args`, but it could happen, that someone uses `args` as normal keyword:
```
threading.Thread(group=None, target=None, name=None, args=(), kwargs={}, *, daemon=None)
scipy.optimize.minimize_scalar(fun, bracket=None, bounds=None, args=(), method='brent', tol=None, options=None)[source]
```
We didn't find a relevant example, but it would be better to prevent the conflict.

[...]

PR #74, opened 2020-11-12
initial commit of Configurable docu

jensheit:
This is an initial commit and further sections will be added. However, I have opened this pull request to already allow for discussions about the first drafts of each section.

### DIFF
--- a/doc/configurable.md
+++ b/doc/configurable.md
@@ -1,0 +1,206 @@
+# Configurable: What is it good for?
+
+The main idea of the configurable object is to simplyfy the initialization
+of a neural network model from a config dictionary.
+A neural network usually consists of multiple modules with inter-dependent
+configurations which are combined in the main model.
+An example model taken from the
+[pytorch tutorial](https://pytorch.org/tutorials/beginner/blitz/cifar10_tutorial.html#sphx-glr-beginner-blitz-cifar10-tutorial-py)
+looks like this:
+```python
+import torch.nn as nn
+import torch.nn.functional as F
+
+class Net(nn.Module):
+    def __init__(self):
+        super().__init__()
+        self.conv1 = nn.Conv2d(3, 6, 5)
+        self.pool = nn.MaxPool2d(2, 2)
+        self.conv2 = nn.Conv2d(6, 16, 5)
+        self.fc1 = nn.Linear(16 * 5 * 5, 120)
+        self.fc2 = nn.Linear(120, 84)
+        self.fc3 = nn.Linear(84, 10)
+
+    def forward(self, x):
+        x = self.pool(F.relu(self.conv1(x)))
+        x = self.pool(F.relu(self.conv2(x)))
+        x = x.view(-1, 16 * 5 * 5)
+        x = F.relu(self.fc1(x))
+        x = F.relu(self.fc2(x))
+        x = self.fc3(x)
+        return x
+
+net = Net()
+```
+Instantiating the modules inside of the model makes it more difficult to
+change parameter in the module or leads to an unnecessary high amount of
+input parameters for the model init.
+Therefore, we assume that the already initialized modules are the input
+to the model:
+```python
+import torch.nn as nn
+import torch.nn.functional as F
+
+class Net(nn.Module):
+    def __init__(self, conv1, pool, conv2, fc1, fc2, fc3):
+        super().__init__()
+        self.conv1 = conv1
+        self.pool = pool
+        self.conv2 = conv2
+        self.fc1 = fc1
+        self.fc2 = fc2
+        self.fc3 = fc3
+
+    def forward(self, x):
+        x = self.pool(F.relu(self.conv1(x)))
+        x = self.pool(F.relu(self.conv2(x)))
+        x = x.view(-1, 16 * 5 * 5)
+        x = F.relu(self.fc1(x))
+        x = F.relu(self.fc2(x))
+        x = self.fc3(x)
+        return x
+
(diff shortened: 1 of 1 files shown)


boeddeker:
I like your motivation, why a verbose config is useful, but the example may be a bit too large.

Are you able to incorporate some ideas from
https://docs.google.com/presentation/d/1SKXlj34niGxVlcTnGAt4KTcymKaAfg7KCNYMO4C1Kho/edit#slide=id.g852ae286d5_3_40
?

I think it would be useful to highlight more, how the config is converted to the actual python object.

doc/configurable.md
```diff
@@ -0,0 +1,206 @@
...
+    def forward(self, x):
+        x = self.pool(F.relu(self.conv1(x)))
+        x = self.pool(F.relu(self.conv2(x)))
+        x = x.view(-1, 16 * 5 * 5)
+        x = F.relu(self.fc1(x))
+        x = F.relu(self.fc2(x))
+        x = self.fc3(x)
+        return x
+
+config = dict(
```
boeddeker, line 92: How about calling this `update`?

doc/configurable.md
```diff
@@ -0,0 +1,206 @@
...
```
boeddeker, line 165: How about calling this update?

doc/configurable.md
```diff
@@ -0,0 +1,206 @@
...
+        in_features=120,
+        out_features=84
+    ),
+    fc3=dict(
+        factory=nn.Linear,
+        in_features=84,
+        out_features=10
+    ),
+)
+full_config = Net.get_config(config)
```
boeddeker, line 200: Could you add `pt.Configurable.get_config(config)` as alternative, when `Net` does not inherent from `pt.Module`?

In general `Configurable` is some kind of "duck" type.
As along as the `__init__` (or function) has no positional only argument (in discussion #61) and all arguments are  `Configurable` the class (or function) is `Configurable`.

doc/configurable.md
```diff
@@ -0,0 +1,206 @@
...
+    ),
+)
+net = pt.Configurable.from_config(config)
+```
+This initialization looks more complicated than the two possiblities above,
+but it has major upsites.
+The main benefit is that we can now easily adjust the parameters from outside
+of the model.
+Another benefit is, that this initialization allows to easily save the used
+configuration for example by writting the config dictionary to a json file.
```
boeddeker, line 134: Could you change `json` to either `yaml` or `yaml/json`?
I think for most people it is easier to read and modify yaml.

doc/configurable.md
```diff
@@ -0,0 +1,206 @@
...
+However, the current config does not include all parameters used in the model
+since there are some default values in the modules which are not specified in
+the config dict.
+If Net is a sub-class of Configurable (for example by replacing nn.Module with
```
thequilo, line 138: Backticks around Net

thequilo: and `nn.Module`. Backticks are missing in other places in this document as well.

doc/configurable.md
```diff
@@ -0,0 +1,206 @@
...
+padertorch.Module) we can get a config with all parameters in the model by
+calling ```Net.get_config(config)```:
```
thequilo, line 140: For inline, single backticks are sufficient

doc/configurable.md
```diff
@@ -1,10 +1,10 @@
 # Configurable: What is it good for?
 
-The main idea of the configurable object is to simplyfy the initialization
+The main idea of the `Configurable` object is to simplyfy the initialization
```
TCord, line 3: simplify

doc/configurable.md
```diff
@@ -1,10 +1,10 @@
...
 of a neural network model from a config dictionary.
 A neural network usually consists of multiple modules with inter-dependent
 configurations which are combined in the main model.
-An example model taken from the
+An simple model inspired by the model in
```
TCord, line 7: A simple

jensheit:
> Can you add something like:
> 
> The value of the `factory` can be a class or a function. A `str` is also allowed, when it is a python path to a class or a function (e.g. `torch.nn.Linear`). Supporting `str` allows using human-readable serializations like `yaml` and `json`.

done

doc/configurable.md
```diff
@@ -15,9 +15,13 @@ Here `Configurable` takes the class specified with the `factory` key and
 initializes it using the other items in the dictionary as kwargs.
 Therefore, all parameters not mentioned in the dictionary fallback to the
 defaults.
-In case of nested dictionaries all nested classes with a factory key are first 
-initialized by `Configurable` before using them as input to the upper-level 
-class. 
+In case of nested dictionaries all nested classes with a `factory` key are 
+first initialized by `Configurable` before using them as input to the upper-level 
+class. The value for `factory` may either be class, a function or a string.
```
thequilo, line 20: `Callable` in general?

doc/configurable.md
```diff
@@ -15,9 +15,13 @@ Here `Configurable` takes the class specified with the `factory` key and
...
+If `factory` is string it must be a python path to a class
```
thequilo, line 21: Add an example for a python path, like `model.MyModel`?

doc/configurable.md
boeddeker: I think not everyone is familiar with the name `Callable`. So maybe `Callable (e.g. class or function)`?

doc/configurable.md
boeddeker: Here, I would use `torch.nn.Linear`. So the example fits to the earlier example.

PR #75, opened 2020-11-13
Add better error message if best checkpoint is a file instead of a symlink

TCord:
Modify error message to make error tracing easier if the best checkpoint is a file instead of a symlink.

### DIFF
--- a/padertorch/train/hooks.py
+++ b/padertorch/train/hooks.py
@@ -590,8 +590,10 @@ class ValidationHook(SummaryHook):
         best_ckpt_path = ckpt_dir / self._best_ckpt_name
         if best_ckpt_path.is_symlink():
             best_ckpt_path.unlink()
-        best_ckpt_path.symlink_to(self.ckpt_ranking[0][0])
-
+        try:
+            best_ckpt_path.symlink_to(self.ckpt_ranking[0][0])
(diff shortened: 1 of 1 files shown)


padertorch/train/hooks.py
```diff
@@ -590,8 +590,10 @@ def set_best_symlink(self, ckpt_dir):
         best_ckpt_path = ckpt_dir / self._best_ckpt_name
         if best_ckpt_path.is_symlink():
             best_ckpt_path.unlink()
-        best_ckpt_path.symlink_to(self.ckpt_ranking[0][0])
-
+        try:
+            best_ckpt_path.symlink_to(self.ckpt_ranking[0][0])
+        except FileExistsError:
+            raise FileExistsError('Best checkpoint needs to be a symlink to a checkpoint, not a file!')
```
boeddeker, line 596: Can you make two changes?
 - Add symlink name and target to the exeption msg.
 - You could do `raise FileExistsError(...) from None`. That would hide the previous exception.
    - Alternatively use `raise FileExistsError(...) from e`

TCord: I think the target name is sufficient here because the error comes from the target path being a file instead of a symlink.

boeddeker (approved):

PR #78, opened 2020-11-13
Update readme

michael-kuhlmann:
Suggestions for a more appealing README. Still needs some revisions based on the other ongoing modifications of the documentation.


README.md
```diff
@@ -4,40 +4,119 @@
...
+While these two methods are mandatory, you are free to add any further methods to your models.
+Since a `Module` does not need a `review` method, it can be used as a component of a `Model`.
+
+## How to Integrate your Data and Model with the Trainer
+
+The trainer works with any kind of iterable data loader, e.g., `torch.utils.data.DataLoader` or `lazy_dataset.Dataset`.
+The `train` method expects a data iterator as input which yields the training data.
+Optionally, you can add a validation iterator with `Trainer.register_validation_hook`.
+The data iterator can yield batched features of type `numpy.ndarray` or `torch.Tensor` or dictionaries with entries of different types.
+The trainer calls `padertorch.data.example_to_device` which recursively converts numpy arrays to Tensors and moves all Tensors to the available device.
```
boeddeker, line 116: In #25 it was changed that the trainer does not call `padertorch.data.example_to_device`, instead the trainer calls `example_to_device` from the model and the default implementation calls `padertorch.data.example_to_device`.

Could you adjust that?

README.md
```diff
@@ -4,40 +4,119 @@
...
+- **Forget the training loop**: Our `padertorch.Trainer` takes care of the repetitive training loop and allows you to focus on network tuning. Included are features such as:
+  - Periodically executed validation runs
+  - Automatic checkpointing: The parameters of the model and the state of the trainer are periodically saved. The checkpoint interval and number of total checkpoints are customizable: Keep one, some or all checkpoints. We also keep track of the best checkpoint on the validation data given some metric.
+  - Learning rate scheduling
+  - Backoff: Restore the best checkpoint and change the learning rate if the loss is not decreasing.
+  - [Averaging multiple checkpoints]():
+  - [Hooks](padertorch/train/hooks.py): Extend the basic features of the trainer with your own functionality.
+- **Logging**: As logging backend, we use `TensorboardX` to generate a `tfevents` file that can be visualized from a `tensorboard`. Custom values to be logged can be defined in subclasses of `padertorch.Model`.
+- **Test run**: The trainer has a `test_run` function to train the model for few iterations and test if
     - the model is executable (burn test)
```
thequilo, line 28: The entries in this list start with lower-case "the", the other enumerations start with upper-case letters. This is a different type of enumeration as it is the continuation of a sentence. Here, I think that it sounds better if you write it like this (adding punctuation and "and"):

... test if
 - the model is executable,
 - the validation is deterministic, and
 - the model changes the parameter in the training.

README.md
```diff
@@ -4,40 +4,119 @@
...
     - the validation is deterministic/reproducable
     - the model changes the parameter in the training
-- Hooks:
-  - The hooks are used to extend the basic features of the trainer. Usually the user dont rearly care about the hooks. By default a `SummaryHook`, a `CheckpointHook` and a `StopTrainingHook` is registerd. So the user only need to register a `ValidationHook`
-- Checkpointing:
-  - The parameters of the model and the state of the trainer are periodically saved. The intervall can be specified with the `checkpoint_trigger` (The units are `epoch` and `iteration`)
-- Virtual minibatch:
-  - The `Trainer` usually do not know if the model is trained with a single example or multiple examples (minibatch), because the exaples that are yielded from the dataset are directly forwarded to the model. 
+- **Virtual minibatch**:
+  - The `Trainer` usually does not know if the model is trained with a single example or multiple examples (minibatch), because the examples that are yielded from the dataset are directly forwarded to the model.
```
thequilo, line 32: Link the "Virual minibatch size and multiple GPUs" file here?

README.md
```diff
@@ -4,40 +4,119 @@
...
+$ git clone https://github.com/fgnt/padertorch.git
+$ cd padertorch && pip install -e .
+```
+
+# Getting Started
+
+## A Short Explanation of `padertorch.Module` and `padertorch.Model`
+
+You can build your models upon `padertorch.Module` and `padertorch.Model`.
+Both expect a `forward` method which has the same functionality as the `forward` call of  `torch.nn.Module`: It takes data as input, applies some transformations and returns the network output:
```
thequilo, line 75: comma after transformations (I think)

README.md
```diff
@@ -4,40 +4,119 @@
...
-  - lazy_dataset.Dataset, torch.utils.data.DataLoader and other iterables...
-- Validation:
-  - The `ValidationHook` runs periodically and and logs the validations results.
-- Learning rate decay with backoff:
-  - The `ValidationHook` has also parameters to do a learning rate with backoff.
-- Test run: 
-  - The trainer has a test run function to train the model for few iterations and test if
+- **Fast prototyping**: The trainer and models are all compatible with [sacred](https://github.com/IDSIA/sacred) and easily allow hyperparameter changes over the command line. Check out the [configurable](padertorch/configurable.py) module and the [examples](padertorch/contrib/examples) for how it works.
+- **Easily extensible**: Write your own network modules and models based on `padertorch.Module` and `padertorch.Model`.
+- **Seamless integration**: You provide your own data and model. We provide the trainer to wrap around your model and data. One simple command starts the training: `padertorch.Trainer.train`.
```
jensheit, line 18: I think this gives the wrong impression that train works without initialization. Perhaps just write that calling train starts the training?

README.md
```diff
@@ -4,40 +4,119 @@
...
+- **Forget the training loop**: Our `padertorch.Trainer` takes care of the repetitive training loop and allows you to focus on network tuning. Included are features such as:
+  - Periodically executed validation runs
+  - Automatic checkpointing: The parameters of the model and the state of the trainer are periodically saved. The checkpoint interval and number of total checkpoints are customizable: Keep one, some or all checkpoints. We also keep track of the best checkpoint on the validation data given some metric.
+  - Learning rate scheduling
+  - Backoff: Restore the best checkpoint and change the learning rate if the loss is not decreasing.
+  - [Averaging multiple checkpoints]():
+  - [Hooks](padertorch/train/hooks.py): Extend the basic features of the trainer with your own functionality.
+- **Logging**: As logging backend, we use `TensorboardX` to generate a `tfevents` file that can be visualized from a `tensorboard`. Custom values to be logged can be defined in subclasses of `padertorch.Model`.
```
jensheit, line 26: perhaps add links to tensorboard and tensorboardx?

README.md
thequilo: And fix the title of the "virual" minibatch size section

README.md
```diff
@@ -4,40 +4,119 @@
...
+- **Test run**: The trainer has a `test_run` function to train the model for few iterations and test if
```
jensheit, line 27: last line: during training instead of in training

README.md
```diff
@@ -4,40 +4,119 @@
...
-  - The hooks are used to extend the basic features of the trainer. Usually the user dont rearly care about the hooks. By default a `SummaryHook`, a `CheckpointHook` and a `StopTrainingHook` is registerd. So the user only need to register a `ValidationHook`
-- Checkpointing:
-  - The parameters of the model and the state of the trainer are periodically saved. The intervall can be specified with the `checkpoint_trigger` (The units are `epoch` and `iteration`)
-- Virtual minibatch:
-  - The `Trainer` usually do not know if the model is trained with a single example or multiple examples (minibatch), because the exaples that are yielded from the dataset are directly forwarded to the model. 
+- **Virtual minibatch**:
+  - The `Trainer` usually does not know if the model is trained with a single example or multiple examples (minibatch), because the examples that are yielded from the dataset are directly forwarded to the model.
   - When the `virtual_minibatch_size` option is larger than one, the trainer calls the forward and backward step `virtual_minibatch_size` times before applying the gradients. This increases the minibatch size, while the memory consumption stays similar.
```
jensheit, line 33: perhaps better: Thereby increasing the number of examples over which the gradient is calculated without higher memory consumption

Otherwise it sounds like everything a higher minibatch size is doing is happening here as well.

README.md
```diff
@@ -86,6 +165,9 @@ trainer.register_validation_hook(validation_dataset)
 trainer.train(train_dataset)
 ```
 
-## padertorch.Configurable
+See the [trainer](padertorch/train/trainer.py#L40) for an explanation of its signature.
+If you want to use `pt.io.get_new_storage_dir` to manage your experiments, you have to define an environment variable `STORAGE_ROOT` which points to the path where all your experiments will be stored, i.e., in the example above, a new directory under `$STORAGE_ROOT/my_experiment_1` will be created.
```
jensheit, line 169: I think we should advertise using either get_new_storage_dir or use get_new_subdir directly because in the second case you do not have to define STORAGE_ROOT and can still use all other benefits.

README.md
```diff
@@ -86,6 +165,9 @@ trainer.register_validation_hook(validation_dataset)
...
+
+## Further Reading
 
-ToDo
+Have a look at our [examples](padertorch/contrib/examples) for advanced usage of padertorch with actual data and models.
```
jensheit, line 173: This should later on direct to the README in examples

jensheit (commented):
This looks much better than the earlier version.

README.md
```diff
@@ -4,40 +4,119 @@
...
+Typically, the input to the `forward` of a `Module` is a Tensor, while for a `Model`, it is a dictionary which contains additional entries, e.g., labels, which are needed in the `review`.
+This is only a recommendation and there is no restriction for the input type.
+
+While these two methods are mandatory, you are free to add any further methods to your models.
+Since a `Module` does not need a `review` method, it can be used as a component of a `Model`.
+
+## How to Integrate your Data and Model with the Trainer
+
+The trainer works with any kind of iterable data loader, e.g., `torch.utils.data.DataLoader` or `lazy_dataset.Dataset`.
+The `train` method expects a data iterator as input which yields the training data.
```
TCord, line 113: Is iterator the correct word, here? I seem to remember we wanted to use "iterable objects" instead, as the dataset does not exactly iterate.

README.md
```diff
@@ -4,40 +4,119 @@
...
+
+# Features for Application in Deep Learning
+
+Padertorch provides a selection of frequently used network architectures and functionalities such as activation and normalization, ready for you to integrate into your own models.
+
+- [Multi-layer Feed-Forward](padertorch/modules/fully_connected.py): Multiple fully-connected layers with non-linearity and dropout.
+- [CNN](padertorch/contrib/je/modules/conv.py) (currently subject to breaking changes and hardly any documentation): 1d- and 2d-CNNs with residual connections, dropout, gated activations, batch and sequence norm and correct handling of down- and upsampling.
+- [Normalization](padertorch/modules/normalization.py): Perform normalization of arbitrary axes/dimensions of your features, keep track of running statistics and apply learnable affine transformation.
+- [Collection of activation functions](padertorch/ops/mappings.py): Fast access of various activation functions with just a string.
+- [Losses](padertorch/ops/losses): We provide an implementation of the [Kullback-Leibler divergence](padertorch/ops/losses/kl_divergence.py) and different [regression](padertorch/ops/losses/regression.py) objectives.
```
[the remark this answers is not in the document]
michael-kuhlmann: I mentioned it below under source separation objectives since I do not find it general enough to be mentioned here.

README.md
boeddeker: `iterator` is the wrong name, that was the initial name that we used for `lazy_dataset`.
 Here, an idea from my side:
```
The trainer works with any kind of iterable, e.g., `list`,
`lazy_dataset.Dataset` or `torch.utils.data.DataLoader`.
The `train` method expects an iterable as input which
yields examples or minibatches of examples that are
forwarded to the model (i.e., the trainer does not interpret
the yielded entries of the iterable, in our examples it
is always a `dict`).
```

michael-kuhlmann: Yes, iterable was also used in the previous version of the README. I will update that.

README.md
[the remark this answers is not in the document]
TCord: Yes, I also noticed that after my remark. In my opinion, it is not prominent enough there, though.

README.md
```diff
@@ -4,40 +4,124 @@
...
+
+## Advanced Architectures
+
+- [Dual-Path RNN (DPRNN)](padertorch/modules/dual_path_rnn.py): See the [paper](https://arxiv.org/abs/1910.06379).
```
jensheit, line 49: Perhaps mention wavenet here as well? I think it is at least as prominent

jensheit:
Could we update the visual style by adding some kind of picture? For example adding a screen shot of tensorboard for one of our examples? Or a block diagram of the default workflow for a training and evaluation? I am not yet sure whether that would improve the look. What are your thoughts?

README.md
michael-kuhlmann: Good idea. I am unsure whether to add it here or below the next section since the Wavenet is more relevant for speech.
I am also thinking of removing this section and putting the DPRNN below the next section, too.

jensheit: I think we can move all of this further down, since these are benefits but not the main selling points.

README.md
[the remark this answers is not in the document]
michael-kuhlmann: Do you have any suggestions on how to improve that? I don't actually know what to add there.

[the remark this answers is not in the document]
TCord: I think removing the point on PIT further down and simply adding a sentence that we also provide a loss wrapper for permutation invariant training criteria would be the best solution.

michael-kuhlmann:
> Could we update the visual style by adding some kind of picture? For example adding a screen shot of tensorboard for one of our examples? Or a block diagram of the default workflow for a training and evaluation? I am not yet sure whether that would improve the look. What are your thoughts?

It is worth it when it can emphasize functionalities of padertorch that are not easily described in words. E.g., pytorch-lightning did quite a good job with their little video. Maybe adding a GIF / short video that shows the call of an example on the command line with the sacred configuration, test_run, and first training steps and then subsequent calls with changed hyperparameters?

TCord:
> > Could we update the visual style by adding some kind of picture? For example adding a screen shot of tensorboard for one of our examples? Or a block diagram of the default workflow for a training and evaluation? I am not yet sure whether that would improve the look. What are your thoughts?
> 
> It is worth it when it can emphasize functionalities of padertorch that are not easily described in words. E.g., pytorch-lightning did quite a good job with their little video. Maybe adding a GIF / short video that shows the call of an example on the command line with the sacred configuration, test_run, and first training steps and then subsequent calls with changed hyperparameters?

One thing, I can think of to convey a better understanding of Padertorch might be a block diagram, where Padertorch frames the processing steps it makes more simple and indicates how the several components interact with each other. Do you think that might be a helpful visualization for the starting page?

README.md
```diff
@@ -4,40 +4,116 @@
...
+```python
+def forward(self, example):
+    x = example['x']
+    out = transform(x)
+    return out
+```
+
+Additionally, `padertorch.Model` expects a `review` method to be implemented which takes the data and output of the `forward` call as inputs from which it computes the training loss and metrics for logging in tensorboard:
+
+```python
```
thequilo, line 78: I think (not sure if this is really the case for everyone) that it is easier to understand if you include the `class MyModel(pt.Model)` part like

```python
class MyModel(pt.Model):
    def forward(self, example):
        x = example['x']
        out = transform(x)
        return out
```
for the method examples.

README.md
```diff
@@ -86,6 +162,36 @@ trainer.register_validation_hook(validation_dataset)
 trainer.train(train_dataset)
 ```
 
-## padertorch.Configurable
+See the [trainer](padertorch/train/trainer.py#L40) for an explanation of its signature.
+If you want to use `pt.io.get_new_storage_dir` to manage your experiments, you have to define an environment variable `STORAGE_ROOT` which points to the path where all your experiments will be stored, i.e., in the example above, a new directory under `$STORAGE_ROOT/my_experiment_1` will be created.
```
thequilo, line 166: Is this correct? I think it creates `$STORAGE_ROOT/my_experiment/1`.

README.md
```diff
@@ -86,6 +162,36 @@ trainer.register_validation_hook(validation_dataset)
...
+
+# Features for Application in Deep Learning
+
+Padertorch provides a selection of frequently used network architectures and functionalities such as activation and normalization, ready for you to integrate into your own models.
+
+- [Multi-layer Feed-Forward](padertorch/modules/fully_connected.py): Multiple fully-connected layers with non-linearity and dropout.
+- [CNN](padertorch/contrib/je/modules/conv.py) (currently subject to breaking changes and hardly any documentation): 1d- and 2d-CNNs with residual connections, dropout, gated activations, batch and sequence norm and correct handling of down- and upsampling.
+- [Normalization](padertorch/modules/normalization.py): Perform normalization of arbitrary axes/dimensions of your features, keep track of running statistics and apply learnable affine transformation.
+- [Collection of activation functions](padertorch/ops/mappings.py): Fast access of various activation functions with just a string.
+- [Losses](padertorch/ops/losses): We provide an implementation of the [Kullback-Leibler divergence](padertorch/ops/losses/kl_divergence.py) and different [regression](padertorch/ops/losses/regression.py) objectives.
```
thequilo, line 177: Should we add examples of the regression losses? Something like "and different regressions objectives (SI-SNR, log-MSE and others)". Many people coming here for source separation probably want to use SI-SNR and won't know that we classified them as regression losses.

README.md
```diff
@@ -86,6 +162,36 @@ trainer.register_validation_hook(validation_dataset)
...
+
+## Support for Sequential and Speech Data
+
+Padertorch especially offers support for training with [sequential data](padertorch/ops/sequence) such as:
+- [Masking](padertorch/ops/sequence/mask.py): Calculate a mask which has non-zero entries for non-padded positions and zero entries for padded positions in the sequence.
+- [Fragmenting](padertorch/data/fragmenter.py): Fragment a sequence into smaller chunks of the same length.
```
thequilo, line 183: -> Segmenting, Segmenter

README.md
```diff
@@ -4,40 +4,116 @@
...
+  - einops
+  - tqdm
+  - natsort
+  - [lazy_dataset](https://github.com/fgnt/lazy_dataset)
+  - IPython
+  - [paderbox](https://github.com/fgnt/paderbox)
+
+```bash
+$ git clone https://github.com/fgnt/padertorch.git
+$ cd padertorch && pip install -e .
```
boeddeker, line 59: Can you move this above `Requirements` and change `pip install -e .` to `pip install -e .[all]`?
Below that you may write, `This will install all dependencies, for a light installation you can drop "[all]"`

README.md
```diff
@@ -67,27 +69,31 @@ You can build your models upon `padertorch.Module` and `padertorch.Model`.
 Both expect a `forward` method which has the same functionality as the `forward` call of  `torch.nn.Module`: It takes data as input, applies some transformations, and returns the network output:
 
 ```python
-def forward(self, example):
-    x = example['x']
-    out = transform(x)
-    return out
+class MyModel(pt.Model):
```
thequilo, line 72: Use `pt.Module` here because the implementation here doesn't have a `review`

README.md
```diff
@@ -67,27 +69,31 @@ You can build your models upon `padertorch.Module` and `padertorch.Model`.
...
+
+  def forward(self, example):
+      x = example['x']
+      out = transform(x)
+      return out
 ```
 
 Additionally, `padertorch.Model` expects a `review` method to be implemented which takes the data and output of the `forward` call as inputs from which it computes the training loss and metrics for logging in tensorboard:
```
thequilo, line 80: I would add something like "This is an example for a classification problem using the cross-entropy loss"

README.md
```diff
@@ -67,27 +69,31 @@ You can build your models upon `padertorch.Module` and `padertorch.Model`.
...
-    # loss computation
-    ce_loss = torch.nn.CrossEntropyLoss()(output, example['label'])
-    # compute additional metrics
-    with torch.no_grad():
-        prediction = torch.argmax(output, dim=1)
-        accuracy = (prediction == example['label']).float().mean()
-    return {
-        'loss': ce_loss,
-        'scalars': {'accuracy': accuracy}
+class MyModel(pt.Model):
```
boeddeker, line 85: Add a dummy `forward` with hidden code (e.g. `...` as function body)?
```python
def forward(self, example):
    ...
```

jensheit (approved):

TCord (approved):

boeddeker (approved):

thequilo (approved):

PR #79, opened 2020-11-16
Add "create_snapshot" flag to model

thequilo:


padertorch/base.py
```diff
@@ -253,6 +256,9 @@ def review(self, inputs, outputs):
          - The contextmanager `torch.no_grad()` disables backpropagation for
            metric computations (i.e. scalars in tensorboard)
          - `self.training` (bool) indicate training or validation mode
+         - `self.create_snapshot` (bool) indicates when to create a snapshot.
+            Useful for reporting of things that need many computational
+            resources (e.g., istft)
```
boeddeker, line 261: Maybe highlighting more, that only one snapshot is stored from the summary hook?
e.g.
> `self.create_snapshot` (bool) indicates when to create a snapshot.
> Motivation: Only one snapshot is reported from the summary hook for snapshot types (e.g. images, audios, ...) for each summary interval. You can use this flag to avoid unnecessary computations and compute only one snapshot that will be reported instead of computing many snapshot that will later be deleted.

jensheit (approved):

padertorch/train/hooks.py
```diff
@@ -375,10 +377,18 @@ def pre_step(self, trainer: 'pt.Trainer'):
             self.finalize_summary(trainer)
             self.dump_summary(trainer)
 
+        # The check using the trigger doesn't work when the hook is loaded from
+        # a snapshot (i.e., set_last got invoked). Because of this, we have a
+        # flag that is set in reset_summary to determine when to compute
+        # snapshots
+        trainer.model.create_snapshot = self.create_snapshot
+        self.create_snapshot = False
```
boeddeker, line 385: Can you change this to something like the following?
```python
if self.create_snapshot:
    trainer.model.create_snapshot = True
    self.create_snapshot = False
```
That would avoid a potential conflict between the hook.
e.g. Assign multiple summary hooks with different report intervals.

tests/test_train/test_hooks.py
```diff
@@ -244,6 +243,29 @@ def review(self, inputs, outputs):
...
+            return super().review(example, output)
+
+    ds = [0., 1., 2.]
+    with tempfile.TemporaryDirectory() as tmp_dir:
+        optimizer = pt.optimizer.Adam()
+        model = Model([1, 2, 3] * 10, tmp_dir, optimizer)
+        trainer = pt.Trainer(
+            model, tmp_dir, optimizer, stop_trigger=(10, 'epoch'),
+            summary_trigger=(1, 'epoch')
+        )
```
boeddeker, line 264: Could you add a second test where you also add `trainer.register_validation_hook`?

padertorch/base.py
thequilo: Done

padertorch/train/hooks.py
thequilo: Done

tests/test_train/test_hooks.py
thequilo: Done

boeddeker (approved):

PR #80, opened 2020-11-16
Add documentation for how to use sacred

thequilo:


doc/sacred.md
```diff
@@ -0,0 +1,175 @@
...
+    configuration values.
+    """
+    # padertorch.Trainer inherits from Configurable, so we can easily construct 
+    # the config dictionary here and instantiate the trainer from the config 
+    # later
+    trainer = {
+        # The 'factory' key tells the Configurable which class to instantiate
+        'factory': pt.Trainer,
+        # The next line creates a new experiment folder for this experiment
+        'storage_dir': pt.io.get_new_storage_dir(ex.path),
```
boeddeker, line 76: The default value of the `mkdir` argument of `pt.io.get_new_storage_dir` is True.
For a resume or cli overwrite it is better, when you do something like:
```python
    trainer = {..., 'storage_dir': None, ...}
    if trainer['storage_dir'] is None:
        trainer['storage_dir'] = pt.io.get_new_storage_dir(ex.path)
```
It looks not so nice, but works better.

doc/sacred.md
```diff
@@ -0,0 +1,175 @@
...
+        Path(trainer['storage_dir']) / 'sacred')
+    )
+
+
+@ex.automain
+def main(trainer, _config):
+    """This is the main function of your experiment. It receives all 
+    configuration values set in `config` as parameters. `_config` contains the  
+    whole configuration as a dictionary."""
+
```
boeddeker, line 103: How about adding `print_config`?

doc/sacred.md
```diff
@@ -0,0 +1,175 @@
...
+    # The following line creates the trainer from the configuration using 
+    # classmethods provided by the Configurable class
+    trainer = pt.Trainer.from_config(trainer)
+
+    # This stores the configuration in the way it is expected by the model 
+    # loading functions in padertorch (pt.Model.from_storage_dir)
+    pt.io.dump_config(_config, trainer.storage_dir / 'config.json')
```
boeddeker, line 110: I think for configs it is better to use yaml.
It is better for humans to read and edit.
The drawback, that it is slower, doesn't matter for so small files.

jensheit: I disagree, we should only promote something we are also using. Most if not all our examples are using json so that it would look strange if we promote something here which we do not heed.

thequilo: And also the default value for `pt.Model.from_storage_dir` is `config.json`, not yaml. We wold have to change that if we recommend to use yaml (or infer the file type)

boeddeker: Good point.
I think using the json comes from the fact, that sacred has json as default and originally, sacred was responsible to write the config.
Maybe I should open an issue and start a discussion, if we should change to yaml as default.

doc/sacred.md
```diff
@@ -0,0 +1,175 @@
...
+    
+    # Run your experiment, start the training or do whatever you want here...
+```
+
+You can run the script with the argument `print_config` to print the configuration it is going to use.
```
TCord, line 115: Change it -> the experiment to make clear what is meant.

PR #81, opened 2020-11-16
move and update wavenet example

JanekEbb:


padertorch/contrib/examples/audio_synthesis/wavenet/data.py
```diff
@@ -0,0 +1,90 @@
+import numpy as np
+from padercontrib.database.librispeech import LibriSpeech
+from padertorch.contrib.je.data.transforms import AudioReader, STFT, \
+    fragment_signal, Collate
+
+
+def get_datasets(audio_reader, stft, max_length, batch_size, shuffle_test=False):
+    db = LibriSpeech()
```
jensheit, line 8: can you change this so you are not using the LibriSpeech Database object but the lazy.database.JsonDatabase with a json_path which is preset to the NT_DATABASE_JSON_PATH / librispeech.json?

padertorch/contrib/examples/audio_synthesis/wavenet/evaluate.py
```diff
@@ -0,0 +1,119 @@
...
+from paderbox.io.new_subdir import get_new_subdir
+from paderbox.io import load_json, dump_json
+
+
+ex = Experiment('wavenet-eval')
+
+
+@ex.config
+def config():
+    exp_dir = ''
```
jensheit, line 27: Here you should use the get_new_storage_dir function from padertorch

jensheit: and probably call it storage_dir and add a storage observer from sacred

padertorch/contrib/examples/audio_synthesis/wavenet/train.py
```diff
@@ -0,0 +1,94 @@
...
+):
+    commands.print_config(_run)
+    trainer = Trainer.from_config(trainer)
+    storage_dir = Path(trainer.storage_dir)
+    storage_dir.mkdir(parents=True, exist_ok=True)
+    commands.save_config(
+        _run.config, _log, config_filename=str(storage_dir / 'config.json')
+    )
+
+    db = LibriSpeech()
```
jensheit, line 80: here holds the same as for the prepare_data, it is better to use lazy_dataset.database to be independent of padercontrib

padertorch/contrib/examples/audio_synthesis/wavenet/evaluate.py
JanekEbb: exp_dir is required to be inputted by the user which is the path to the experiment directory with the model in it that is to be evaluated. get_new_subdir is used in line 40.

padertorch/contrib/examples/audio_synthesis/wavenet/train.py
```diff
@@ -0,0 +1,94 @@
...
+from padertorch.train.trainer import Trainer
+from sacred import Experiment, commands
+
+ex = Experiment('wavenet')
+
+
+@ex.config
+def config():
+    training_sets = ['train_clean_100', 'train_clean_360', 'train_other_500']
+    validation_sets = ['dev_clean', 'dev_other']
```
michael-kuhlmann, line 23: In the README, you wrote that it was trained on train_clean_100 + train_clean_360. Is it reasonable to also train on noisy data? The most use cases I know of are concerned with clean speech.

padertorch/contrib/examples/audio_synthesis/wavenet/README.md
```diff
@@ -0,0 +1,56 @@
...
+extraction and normalization is done in the model.
+
+## Evaluation
+
+The evaluation script loads the best checkpoint (lowest achieved loss on the 
+validation set) and performs autoregressive waveform synthesis.
+For test-time synthesis nv-wavenet needs to be installed.
+Do note that nv-wavenet requires Compute Capability 6.0 or later 
+(https://developer.nvidia.com/cuda-gpus), i.e., you can not, e.g., run the 
+evaluation on a GTX 980.
```
michael-kuhlmann, line 24: Maybe also mention that it does not run on a CPU?

michael-kuhlmann: I think "cannot" is better here

padertorch/contrib/examples/audio_synthesis/wavenet/train.py
JanekEbb: In the README I show results when training only with clean (which however are from an old model and may be updated at some point). I will add results when training on all data as soon as I have them. Do you think it is disadvantageous to add noisy data in the training?

michael-kuhlmann: Not necessarily disadvantageous, but just rarely seen in cases where WaveNets are applied like TTS and VC (which use even cleaner data than LibriSpeech clean for training). If you report the results of both experiments (with and without train_other_500), I have no complaints about this.

jensheit (approved):

michael-kuhlmann (approved):

PR #82, opened 2020-11-16
Segment functions for Segmenter

jensheit:
This pull request adds segment and get_segment_offsets which can be used in a Segmenter.

### DIFF
--- a/padertorch/data/fragmenter.py
+++ b/padertorch/data/fragmenter.py
@@ -128,7 +128,8 @@ def get_anchor(
     """
     Calculates anchor for the boundaries for segmentation of a signal
     with length `num_sammples` in case of a fixed segment length
-    `length` and shift `shift`
+    `length` and shift `shift`. The anchor always points to the first value of
+    a segment.
 
     Args:
         num_samples: num samples of signal for which boundaries are caclulated
--- a/padertorch/data/fragmenter.py
+++ b/padertorch/data/fragmenter.py
@@ -109,13 +109,14 @@ class Fragmenter(object):
         return fragments
 
 
-def get_segment_offsets(num_samples, length, step=None,
+
+def get_segment_offsets(num_samples, length, shift=None,
                         offset_mode='start', num_segments=None):
     """
 
     Args:
         num_samples:
-        segment_length:
+        length:
         shift:
         offset_mode:
         num_segments:
@@ -137,31 +138,31 @@ def get_segment_offsets(num_samples, length, step=None,
     array([1])
     """
     assert num_samples >= length, (num_samples, length)
-    if step is None:
-        step = length
-    assert step > 0, step
+    if shift is None:
+        shift = length
+    assert shift > 0, shift
 
     if offset_mode == 'start':
         start = 0
     elif offset_mode == 'random':
         start = np.random.randint(num_samples - length)
     elif offset_mode in ['end', 'center']:
-        remainder = (num_samples - length) % step
+        remainder = (num_samples - length) % shift
         if offset_mode == 'end':
             start = remainder
         else:
             start = remainder // 2
     else:
         raise ValueError('Unknown offset mode', offset_mode)
-    offsets = np.arange(start, num_samples - length + 1, step)
+    offsets = np.arange(start, num_samples - length + 1, shift)
     if num_segments:
-        max_segments = (num_samples - length) // step + 1
+        max_segments = (num_samples - length) // shift + 1
         assert num_segments <= max_segments, (num_segments, max_segments)
         offsets = offsets[:num_segments]
-    return offsets
+    return  offsets
 
 
-def segment(x, length, step=None, axis=-1,
+def segment(x, length, offsets=None, shift=None, axis=-1,
             offset_mode='start', num_segments=None):
     """
 
@@ -179,15 +180,14 @@ def segment(x, length, step=None, axis=-1,
     array([[ 0,  1,  2,  3,  4,  5,  6,  7,  8,  9],
            [ 3,  4,  5,  6,  7,  8,  9, 10, 11, 12]])
     """
-    ndim = x.ndim
-    axis = axis % ndim
 
-    offsets = get_segment_offsets(x.shape[axis], length, step,
-                                  offset_mode, num_segments)
+    if offsets is None:
+        offsets = get_segment_offsets(x.shape[axis], length, shift,
+                                      offset_mode, num_segments)
     # slice the array to remove samples not addressed with
     # this offsets-segment_length combination
-    slc = [slice(None)] * ndim
+    slc = [slice(None)] * x.ndim
     slc[axis] = slice(offsets[0], offsets[-1] + length)
     x = x[tuple(slc)]
 
-    return segment_axis(x, length, step, end='cut')
+    return segment_axis(x, length, shift, end='cut')
--- a/padertorch/data/fragmenter.py
+++ b/padertorch/data/fragmenter.py
@@ -212,9 +212,9 @@ def segment(
 
     Args:
         x: signal to be segmented
-        boundary: 2xB numpy array or list with start and end values
-            for B pre calculated boundaries, which may be used to segment
-            multiple audios in the same way.
+        boundary: 1xB or 2xB `numpy array`, `tuple` or `list` containing start
+            (and end) values for B pre calculated boundaries.
+            If only start values are given the segment length has to be set.
         length: segment length
         axis: axis which is segmented
         shift: shift between segments, defaults to length
@@ -268,8 +268,13 @@ def segment(
             # This might be necessary if only one start and stop point is given
             boundary = boundary[..., None]
 
-    assert boundary.shape[0] == 2, 'The first dimension has to be 2 to ' \
-                                   'describe start and stop of a segment'
+    if boundary.shape[0] == 1 and length is not None:
+        boundary = np.concatenate((boundary, boundary + length), axis=0)
+
+    assert boundary.shape[0] == 2, ('The first dimension of boundary has to ' \
+                                    'be  2 to describe start and stop of a ' \
+                                    'segment or can be 1 if a fixed length ' \
+                                    'is defined', boundary.shape, length)
     assert boundary.ndim == 2, 'Expects an two dimensional boundary array'
 
     if boundary.shape[1] == 1:
--- a/padertorch/data/fragmenter.py
+++ b/padertorch/data/fragmenter.py
@@ -122,7 +122,7 @@ class Fragmenter(object):
 
 
 def get_anchor(
-        num_samples: int, length: int, shift: int=None,
+        num_samples: int, length: int, shift: int = None,
         mode: str = 'left', rng=np.random
 ):
     """
@@ -192,42 +192,43 @@ def get_anchor(
 
 
 def get_segment_boundaries(
-        anchor: Union[str, int], num_samples: int, length: int,
-        shift: int = None, rng=np.random
+        num_samples: int, length: int, shift: int = None,
+        anchor: Union[str, int] = 'left', rng=np.random
 ):
     """
     Calculates boundaries for segmentation of a signal with length
-    `num_sammples` in case of a fixed segment length `length` and shift `shift`
+    `num_samples` in case of a fixed segment length `length` and shift `shift`
 
     Args:
-        anchor: anchor from which the segmentation boundaries are calculated.
-            if it is a string `get_anchor` is called to calculate an integer
-            using `anchor` as anchor mode definition.
-        num_samples: num samples of signal for which boundaries are caclulated
+        num_samples: number of samples of signal for which
+            boundaries are calculated.
         length: segment length
         shift: shift between segments, defaults to length
+        anchor: anchor from which the segmentation boundaries are calculated.
+            If it is a string `get_anchor` is called to calculate an integer
+            using `anchor` as anchor mode definition.
         rng: random number generator (`numpy.random`)
 
     Returns:
         Bx2 numpy array with start and end values for B boundaries
 
     >>> np.random.seed(3)
-    >>> get_segment_boundaries('left', 24, 10, 3).T
+    >>> get_segment_boundaries(24, 10, 3, anchor='left').T
     array([[ 0,  3,  6,  9, 12],
            [10, 13, 16, 19, 22]])
-    >>> get_segment_boundaries('right', 24, 10, 3).T
+    >>> get_segment_boundaries(24, 10, 3, anchor='right').T
     array([[ 2,  5,  8, 11, 14],
            [12, 15, 18, 21, 24]])
-    >>> get_segment_boundaries('center', 24, 10, 3).T
+    >>> get_segment_boundaries(24, 10, 3, anchor='center').T
     array([[ 0,  3,  6,  9, 12],
            [10, 13, 16, 19, 22]])
-    >>> get_segment_boundaries('centered_cutout', 24, 10, 3).T
+    >>> get_segment_boundaries(24, 10, 3, anchor='centered_cutout').T
     array([[ 1,  4,  7, 10, 13],
            [11, 14, 17, 20, 23]])
-    >>> get_segment_boundaries('random', 24, 10, 3).T
+    >>> get_segment_boundaries(24, 10, 3, anchor='random').T
     array([[ 1,  4,  7, 10, 13],
            [11, 14, 17, 20, 23]])
-    >>> get_segment_boundaries('random_max_segments', 24, 10, 3).T
+    >>> get_segment_boundaries(24, 10, 3, anchor='random_max_segments').T
     array([[ 0,  3,  6,  9, 12],
            [10, 13, 16, 19, 22]])
     """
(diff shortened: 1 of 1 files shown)


padertorch/data/fragmenter.py
```diff
@@ -107,3 +107,95 @@ def fragment(key, x):
             fragment = deflatten(fragment)
             fragments.append(fragment)
         return fragments
+
+
+def get_segment_offsets(num_samples, segment_length, step=None,
```
boeddeker, line 112: How about to sync the names with the stft?

jensheit: You mean window_length and shift? Sure, although I might go with length and shift.

padertorch/data/fragmenter.py
```diff
@@ -107,3 +107,95 @@ def fragment(key, x):
...
+    array([ 0, 10])
+    >>> get_segment_offsets(25, 10, 3, offset_mode='start')
+    array([ 0,  3,  6,  9, 12, 15])
+    >>> get_segment_offsets(24, 10, 3, offset_mode='start')
+    array([ 0,  3,  6,  9, 12])
+    >>> get_segment_offsets(24, 10, 3, offset_mode='end')
+    array([ 2,  5,  8, 11, 14])
+    >>> get_segment_offsets(24, 10, 3, offset_mode='center')
+    array([ 1,  4,  7, 10, 13])
+    >>> get_segment_offsets(24, 10, 3, offset_mode='center', num_segments=2)
```
boeddeker, line 134: What is the advantage of using `num_segments` compared to `get_segment_offsets(...)[:num_segments]`?

padertorch/data/fragmenter.py
```diff
@@ -107,3 +107,95 @@ def fragment(key, x):
...
+    array([1, 4])
+    >>> get_segment_offsets(24, 10, 3, offset_mode='center', num_segments=1)
+    array([1])
+    """
+    assert num_samples >= segment_length, (num_samples, segment_length)
+    assert step > 0, step
+    if step is None:
+        step = segment_length
```
boeddeker, line 142: move the assert `assert step > 0, step` below this?

padertorch/data/fragmenter.py
jensheit: At the moment there is none. I was thinking about making it random. However, I was not sure whether anybody was interested in that. I am not using this num_segments feature so I am not sure what the expectations are.

padertorch/data/fragmenter.py
```diff
@@ -107,3 +107,88 @@ def fragment(key, x):
...
+            offset_mode='start', num_segments=None):
+    """
+
+    Args:
+        x:
+        segment_length:
+        step:
+        axis:
+        offset_mode:
+        num_segments:
```
thequilo, line 176: The args in the docstring don't match the actual arg names. Also, type annotations and descriptions of the parameters would be nice to have

padertorch/data/fragmenter.py
```diff
@@ -107,3 +107,88 @@ def fragment(key, x):
...
+    Returns:
+
+    >>> segment(np.arange(0, 15), 10, None, 3, -1)
+    array([[ 0,  1,  2,  3,  4,  5,  6,  7,  8,  9],
+           [ 3,  4,  5,  6,  7,  8,  9, 10, 11, 12]])
+    """
+
+    if offset is None:
+        offset = get_segment_offsets(x.shape[axis], length, shift,
+                                      offset_mode, num_segments)
```
thequilo, line 187: Is it necessary to call get_segment_offsets here or would it be enough to move the "offset_mode" part to a separate function and use that function here?

padertorch/data/fragmenter.py
```diff
@@ -109,86 +119,145 @@ def fragment(key, x):
...
 
     Args:
-        num_samples:
-        length:
-        shift:
-        offset_mode:
-        num_segments:
+        num_samples: num samples of signal for which boundaries are caclulated
+        length: segment length
+        shift: shift between segments
```
boeddeker, line 129: add `default is length`?

padertorch/data/fragmenter.py
```diff
@@ -109,86 +119,145 @@ def fragment(key, x):
...
+                                       ' describe start and stop of a segment'
+
     # slice the array to remove samples not addressed with
     # this offsets-segment_length combination
     slc = [slice(None)] * x.ndim
-    slc[axis] = slice(offset[0], offset[-1] + length)
+    slc[axis] = slice(boundary[0, 0], boundary[1, -1])
     x = x[tuple(slc)]
 
     return segment_axis(x, length, shift, end='cut')
```
thequilo, line 263: This only works if `boundaries` has the correct format. If you pass in custom boundaries that don't match up with a constant integer shift and size, this gives unexpected results. Can we also support passing in boundaries that don't use a constant shift and size?

padertorch/data/fragmenter.py
```diff
@@ -109,86 +119,145 @@ def fragment(key, x):
         return fragments
 
 
-
-def get_segment_offsets(num_samples, length, shift=None,
-                        offset_mode='start', num_segments=None):
+def get_segment_boundaries(num_samples, length, shift=None,
+                           mode='begin', choose_one=False, rng=np.random):
```
thequilo, line 123: Add type annotations

padertorch/data/fragmenter.py
```diff
@@ -109,86 +119,145 @@ def fragment(key, x):
...
+        start = rng.choice(start, 1)
+    stop = start + length
+    boundaries = np.stack([start, stop], axis=0)
+    return boundaries
 
 
-def segment(x, length, offset=None, shift=None, axis=-1,
-            offset_mode='start', num_segments=None):
+def segment(x, axis=-1, boundary=None, length: int=None, shift=None,
+            mode='begin', choose_one=False, rng=np.random):
```
thequilo, line 199: Add type annotations for all arguments and return value

padertorch/data/fragmenter.py
```diff
@@ -109,86 +119,145 @@ def fragment(key, x):
         return fragments
 
 
-
-def get_segment_offsets(num_samples, length, shift=None,
-                        offset_mode='start', num_segments=None):
+def get_segment_boundaries(num_samples, length, shift=None,
+                           mode='begin', choose_one=False, rng=np.random):
     """
```
thequilo, line 124: Add a description of this function at the beginning of the function. (What does this function do? best in one sentence)

padertorch/data/fragmenter.py
```diff
@@ -109,86 +119,145 @@ def fragment(key, x):
...
+    boundaries = np.stack([start, stop], axis=0)
+    return boundaries
 
 
-def segment(x, length, offset=None, shift=None, axis=-1,
-            offset_mode='start', num_segments=None):
+def segment(x, axis=-1, boundary=None, length: int=None, shift=None,
+            mode='begin', choose_one=False, rng=np.random):
     """
+    Segments a signal x along an axis. Either with predefined segment
```
thequilo, line 201: Use backticks when referencing arguments in the description text. This looks way nicer when compiled to HTML.

padertorch/data/fragmenter.py
```diff
@@ -109,86 +119,145 @@ def fragment(key, x):
...
+            end: boundaries end at num_samples
+                so that only values at the beginning are cut
+            center: boundaries are chosen such that the same number of samples
+                are discarded at the end and the beginning of the signal
+            random: Starts the first segment randomly at a value
+                between 0 and shift
+            random_max_segments: Randomly chooses the first segment such that
+                the maximum number of segments are created
+        choose_one: If True only outputs one segment boundary
+        rng: random number generator (numpy.random)
```
thequilo, line 225: Backticks for `numpy.random`, `shift`, ... all arguemnts references in the text

thequilo (commented):
Since we want to improve the documentation I added some comments regarding the type annotations and docstrings.

padertorch/data/fragmenter.py
jensheit: good idea, I will replace this with a max and min operator

boeddeker: I think it would be better to drop the end value of boundary and only use `begin` as argument. That is enough and much easier for the user.

boeddeker: If you keep this signature, a user may think, this function can do much more fancy things.

padertorch/data/fragmenter.py
```diff
@@ -107,3 +117,147 @@ def fragment(key, x):
...
+    >>> get_segment_boundaries(24, 10, 3, mode='center', choose_one=True)
+    array([[ 1],
+           [11]])
+    """
+    assert num_samples >= length, (num_samples, length)
+    if shift is None:
+        shift = length
+    assert shift > 0, shift
+
+    if mode == 'begin':
```
JanekEbb, line 180: I think the terms are still not very clear yet. Can we come up with a better name for mode? Maybe boundary_mode? Also begin (btw I think begin is a verb while beginning is the corresponding noun) is ambiguous as it may refer to the beginning of the segments (according to Christoph's comment) or to the beginning of the input. Maybe it is better to call it boundary_mode="flush-left" and  boundary_mode="flush-right" instead of "begin" and "end", respectively. What do the others think? @boeddeker @thequilo

padertorch/data/fragmenter.py
jensheit: hmm, okay I have gone a different way and have added the possiblity to use more possible inputs but you are right I will allow the user to input a one dimensional array and in that case the end part of the array will be added by using a fixed length if it was specified.

padertorch/data/fragmenter.py
jensheit: done

padertorch/data/fragmenter.py
jensheit: done

padertorch/data/fragmenter.py
jensheit: done

padertorch/data/fragmenter.py
jensheit: done

padertorch/data/fragmenter.py
jensheit: done

padertorch/data/fragmenter.py
JanekEbb: support boundaries that don't use a constant shift and size would be cool. How about inferring shifts and lengths from the boundary array? If they are the same for all segments use segment_axis and else a for loop?

padertorch/data/fragmenter.py
thequilo: You are right, "begin" and "end" are verbs and it can be confusing if we use these. Same for start/stop: I don't know what I would expect of the boundary mode "stop". I like the naming with "flush", but I'm not so sure about left and right. I'm fine with them, but is it always clear what left and right mean in this context?

padertorch/data/fragmenter.py
boeddeker: I think there are a few signatures a user may expect:
 - `segment(x, length, shift, mode) -> array`
 - `segment(x, length, shift, begin, end=None) -> array`
 - `segment(x, boundaries) -> list[array]`

When a user specifies the `boundaries` argument, `length` and `shift` should be ignored and the code uses a loopy implementation (i.e. return type is list of array).

When `boundaries` is not given, the return type can be a multidimensional array and `segment_axis` should be used.

To force a user that want to use beginning has to specify boundaries is not a good idea (Too much boilerplate).

I don't like the idea to use magic to figure out if boundaries has common size and shifts.
It can randomly change the return type.

padertorch/data/fragmenter.py
boeddeker: I agree, flush is more obvious. Is `flush-begin` something resonable?

padertorch/data/fragmenter.py
```diff
@@ -107,3 +119,190 @@ def fragment(key, x):
...
+                so that only values at the end are cut
+            right: anchor is set to `num_samples`
+                so that only values at the beginning are cut
+            center: anchor is set to `num_samples // 2`
+            centered_cutout: the anchor is chosen such that the same number
+                of samples are discarded at the end and the beginning
+            random: anchor is set to a random value between
+                `0` and `num_samples`.
+                 This may reduce the number of possible segments.
+            random_max_segments: Randomly chooses the anchor such that
```
boeddeker, line 148: Given the name `centered_cutout`, would then the name `random_cutout` better match than `random_max_segments`?

padertorch/data/fragmenter.py
```diff
@@ -107,3 +119,190 @@ def fragment(key, x):
...
+        return 0
+    elif mode == 'right':
+        return num_samples - length
+    elif mode == 'center':
+        return num_samples // 2
+    elif mode == 'centered_cutout':
+        remainder = (num_samples - length) % shift
+        return remainder // 2
+    elif mode == 'random':
+        return rng.randint(num_samples - length)
```
boeddeker, line 183: Why `num_samples - length`? Shouldn't this valu be in the range `[0, ..., num_samples]`?
Note I included the last value.

padertorch/data/fragmenter.py
```diff
@@ -107,3 +119,190 @@ def fragment(key, x):
...
+    elif mode == 'random_max_segments':
+        start = rng.randint((num_samples - length) % shift + 1)
+        anchors = np.arange(start, num_samples - length + 1, shift)
+        return int(np.random.choice(anchors))
```
boeddeker, line 187: Why do you calculate all equivalent anchors and select a random anchor from this?
You can return any of the calculated anchors and the result later be the same.

jensheit: because there might be a usecase for this if you only calculate one segment and not all of them.

boeddeker: Ah, ok. But in that case the user wants to have a left anchor with random.
So this function may not do exactly what the user wants.

padertorch/data/fragmenter.py
```diff
@@ -120,130 +121,161 @@ def fragment(key, x):
...
+def get_anchor(
         num_samples: int, length: int, shift: int=None,
-        mode: str='begin', choose_one: bool=False, rng=np.random
+        mode: str = 'left', rng=np.random
 ):
     """
-    Calculates boundaries for segmentation of a signal with length
-    `num_sammples` in case of a fixed segment length `length` and shift `shift`
+    Calculates anchor for the boundaries for segmentation of a signal
+    with length `num_sammples` in case of a fixed segment length
```
thequilo, line 130: typo "num_sammples"

padertorch/data/fragmenter.py
```diff
@@ -120,130 +121,161 @@ def fragment(key, x):
...
+                         'choose on of', possible_boundary_modes)
+
+
+def get_segment_boundaries(
+        anchor: Union[str, int], num_samples: int, length: int,
+        shift: int = None, rng=np.random
+):
+    """
+    Calculates boundaries for segmentation of a signal with length
+    `num_sammples` in case of a fixed segment length `length` and shift `shift`
```
thequilo, line 200: Full stop at the end of the sentence

padertorch/data/fragmenter.py
```diff
@@ -120,130 +121,161 @@ def fragment(key, x):
...
+
+    Args:
+        anchor: anchor from which the segmentation boundaries are calculated.
+            if it is a string `get_anchor` is called to calculate an integer
```
thequilo, line 204: If should be uppder case here

padertorch/data/fragmenter.py
```diff
@@ -120,130 +121,161 @@ def fragment(key, x):
...
+            using `anchor` as anchor mode definition.
+        num_samples: num samples of signal for which boundaries are caclulated
```
thequilo, line 206: num samples -> number of samples

padertorch/data/fragmenter.py
jensheit: done

padertorch/data/fragmenter.py
jensheit: done

padertorch/data/fragmenter.py
jensheit: done

padertorch/data/fragmenter.py
jensheit: done

padertorch/data/fragmenter.py
```diff
@@ -122,7 +122,7 @@ def fragment(key, x):
 
 
 def get_anchor(
-        num_samples: int, length: int, shift: int=None,
+        num_samples: int, length: int, shift: int = None,
         mode: str = 'left', rng=np.random
```
thequilo, line 126: How about supporting the new numpy 1.19 random number generator `np.random.default_rng()`? (https://numpy.org/doc/stable/reference/random/new-or-different.html) We would need a typecheck for rng and use `integers` instead of `randint`.

boeddeker: Can you explain the advantage of the new random number generator?
They did some API cleanup and changed the internals, but I don't see the short term advantage.
The short term disadvantage is, that we would require at least numpy 1.19.

jensheit: I think the idea is to support both the current and the future version of random by doing a typecheck on rng, so we would only need 1.19 if the old numpy typecheck did not work.

thequilo: As far as I can see, the main advantage is that they use a better random number generator algorithm.

thequilo: And at some point, they are probably going to drop support for the old `RandomState`.

boeddeker (approved):

thequilo (approved):

PR #84, opened 2020-11-17
move and adjust audio_tagging example

JanekEbb:


jensheit (approved):

padertorch/contrib/examples/sound_recognition/audio_tagging/evaluate.py
```diff
@@ -0,0 +1,177 @@
+"""
+Example call:
+
+python -m padertorch.contrib.examples.audio_synthesis.wavenet.evaluate with exp_dir=/path/to/exp_dir
```
jensheit, line 4: this is probably a copy and paste error

PR #86, opened 2020-11-24
Add Segmenter

jensheit:
This pull request replaces the Fragmenter with a new Segmenter class and adds test cases for this class.


padertorch/data/fragmenter.py
```diff
@@ -15,110 +16,199 @@
...
+        shift: shift between segments, defaults to length
+        include_keys: The keys in the passed example dict to segment. They all
+            must have the same size along their specified `axis`, if keys is
+            None the segmentation is applied to all `numpy.arrays`.
+            If a key points to a dictionary the segmentation is applied to all
+            values of this dictionary
+        exclude_keys: This option allows to specify specific keys not to
+            segment. This might be usefull if `include_keys` is `None` or
+            one of the included keys points to a dictionary.
+        copy_keys: If `None` all not values which are not segmented are
```
boeddeker: Default value of `True` sounds more resonable.

padertorch/data/fragmenter.py
```diff
@@ -15,110 +16,199 @@
...
+                 exclude_keys: Union[str, list, tuple] = None,
+                 copy_keys: Union[str, list, tuple] = None,
+                 axis: Union[int, list, tuple, dict] = -1,
+                 anchor_mode: str = 'left', rng=np.random):
+
+        self.include = None if include_keys is None else to_list(include_keys)
+        self.exclude = [] if exclude_keys is None else to_list(exclude_keys)
+        self.length = length
+        if isinstance(axis, (dict, int)):
+            self. axis = axis
```
boeddeker: pep8: `self. axis` -> `self.axis`

padertorch/data/fragmenter.py
```diff
@@ -15,110 +16,199 @@
...
+        elif isinstance(axis, (tuple, list)):
+            self.axis = to_list(axis)
```
boeddeker: `self.axis = to_list(axis)` -> `self.axis = list(axis)`
Not sure, why you make this operation.

padertorch/data/fragmenter.py
```diff
@@ -15,110 +16,199 @@
...
-            [len(features[key]) for key in list(features.keys())]
+            raise TypeError('Unknown type for axis', axis)
+        self.shift = shift
+        self.anchor_mode = anchor_mode
+        self.rng = rng
+        self.copy_keys = None if copy_keys is None else to_list(copy_keys)
+
+    def __call__(self, example):
+        # Shortcut if segmentation is disabled
+        if self.length == -1:
```
boeddeker: Replace this with `if len(self.include) == 0:`

padertorch/data/fragmenter.py
```diff
@@ -15,110 +16,199 @@
...
+        if self.include is None:
+            return [
+                key for key, value in example.items()
+                if not key in self.exclude and isinstance(value, np.ndarray)
+            ]
+        else:
+            return [
+                key for key in example.keys()
+                if not key in self.exclude and
+                any([include_key in key for include_key in self.include])
```
boeddeker: Use `key.startswith(include_key)`. Or better use `sep=None` in flatten and be robust against dots `.` in the keys.

padertorch/data/fragmenter.py
```diff
@@ -15,110 +16,199 @@
...
+            return [self.axis[key] for key in to_segment_keys]
+        elif isinstance(self.axis, list):
+            assert len(self.axis) == len(to_segment_keys), (
+                f'The list for axis does not include a axis for each'
+                f'segment array. axis: {self.axis}, '
+                f'segment array keys {to_segment_keys}'
+            )
+            return self.axis
+        else:
+            raise NotImplementedError('This should never be reached')
```
boeddeker: `TypeError`?

boeddeker: Add the type and value of axis to the error msg

padertorch/data/segment.py
```diff
@@ -0,0 +1,413 @@
...
+            from which the segmentation boundaries are calculated.
+        rng: random number generator (`numpy.random`)
+    """
+
+    def __init__(self, length: int = -1,shift: int = None,
+                 include_keys: Union[str, list, tuple] = None,
+                 exclude_keys: Union[str, list, tuple] = None,
+                 copy_keys: Union[str, list, tuple] = None,
+                 axis: Union[int, list, tuple, dict] = -1,
+                 anchor_mode: str = 'left', rng=np.random):
```
thequilo, line 88: For the rng, does it make sense to allow a mode where the rng is seeded from the example to have reproducibility?

padertorch/data/segment.py
```diff
@@ -0,0 +1,413 @@
...
+            return [
+                key for key, value in example.items()
+                if not key in self.exclude and isinstance(value, np.ndarray)
+            ]
+        else:
+            return [
+                key for key in example.keys()
+                if not key in self.exclude and
+                any([include_key in key for include_key in self.include])
+            ]
```
thequilo, line 191: What if a key is in `self.include` but not in the example? Should the code pass or should it raise an exception?

padertorch/data/fragmenter.py
```diff
@@ -15,110 +16,199 @@
...
+        boundaries, segmented = self.segment(to_segment, to_segment_length,
+                                             axis=axis)
+
+        segmented_examples = list()
+        if self.copy_keys is not None:
+            example = {key: example.pop(key) for key in self.copy_keys}
+        for idx, (start, stop) in enumerate(boundaries):
+            example_copy = copy(example)
+            example_copy.update({to_segment_keys[i]: value[idx]
+                                 for i, value in enumerate(segmented)})
```
boeddeker: Use a dict for `segmented` and `to_segment` to avoid this hack.

padertorch/data/fragmenter.py
```diff
@@ -15,110 +16,199 @@
...
+
+        boundaries, segmented = self.segment(to_segment, to_segment_length,
+                                             axis=axis)
+
+        segmented_examples = list()
+        if self.copy_keys is not None:
+            example = {key: example.pop(key) for key in self.copy_keys}
+        for idx, (start, stop) in enumerate(boundaries):
+            example_copy = copy(example)
+            example_copy.update({to_segment_keys[i]: value[idx]
```
boeddeker: Is it better to connect boundaries and values already in segment?

padertorch/data/segment.py
```diff
@@ -0,0 +1,413 @@
...
+    def get_axis_list(self, to_segment_keys):
+        if isinstance(self.axis, int):
+            return [self.axis] * len(to_segment_keys)
+        elif isinstance(self.axis, dict):
+            assert all([key in self.axis.keys() for key in to_segment_keys]), (
+                f'The dictionary for axis did not include keys for all'
+                f'segment arrays. axis keys: {self.axis.keys()},'
+                f'segment array keys {to_segment_keys}'
+            )
+            return [self.axis[key] for key in to_segment_keys]
```
thequilo, line 202: How about allowing broadcasting (e.g., `{'audio_data': 1, 'vad': 0}` would use axis 1 for everything under `audio_data` and axis 0 for everything under `vad`)?

thequilo (commented):
A general comment: Shall we allow torch Tensors in addition to numpy arrays as well? I don't see a reason against this.

padertorch/data/fragmenter.py
jensheit: removes the need to ask whether it might be a tuple in the rest of the code

padertorch/data/fragmenter.py
jensheit: I think I can add an option for sep to the init, so that everybody can choose for himself, because if I want to use exclude I would not like to specify the key audiorir instead of audio.rir

padertorch/data/fragmenter.py
jensheit: I have not done that because this error should already be cached in the init. Somebody has to overwrite self.axis for this to happen.

padertorch/data/fragmenter.py
jensheit: I thought about that, but I dislike the use of `is True` to check whether it is equal to True this seems more natural with None, but okay, I can change that.

padertorch/data/segment.py
```diff
@@ -0,0 +1,413 @@
...
+            values of this dictionary
+        exclude_keys: This option allows to specify specific keys not to
+            segment. This might be usefull if `include_keys` is `None` or
+            one of the included keys points to a dictionary.
+        copy_keys: If `None` all not values which are not segmented are
+            copied. Otherwise only the specified keys are added to the new
+            dictionary with the segmented signals.
+        axis: axis over which to segment
+        anchor_mode: anchor mode used in `get_anchor` to calculate the anchor
+            from which the segmentation boundaries are calculated.
```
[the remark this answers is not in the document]
jensheit: I thought about that, but I could not think of a usecase.

padertorch/data/segment.py
jensheit: sure, I can do that. However, I though the user would specify the rng in that case.

padertorch/data/segment.py
jensheit: It should probably raise an exception, I can add that.

padertorch/data/fragmenter.py
thequilo: I wouldn't replace it. You might want to disable chunking from the config and then it is much easier to set the length to -1 than to change the include keys.

padertorch/data/segment.py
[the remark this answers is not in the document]
thequilo: I currently can't think of a use-case either, maybe someone wants to remove a short silence in the beginning? Anyway, I just thought it would be nice to have the same interface as `segment` which allows setting an integer.

padertorch/data/segment.py
thequilo: The user can't specify the rng when the `Segmenter` is mapped to a dataset. Then, the `Segmenter` uses the same rng for all examples

padertorch/data/segment.py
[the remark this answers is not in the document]
jensheit: Sure, I can add that it is not much of a change

padertorch/data/segment.py
jensheit: that is True, perhaps this option should be specified during call?

padertorch/data/fragmenter.py
boeddeker: Ah, ok. Nevertheless, please add axis, just in case the code gets modified: `NotImplementedError('This should never be reached', self.axis)`

padertorch/data/segment.py
```diff
@@ -0,0 +1,442 @@
...
+        [32000 63999]
+        >>> ex = {'x': np.arange(63999), 'y': np.arange(63999)}
+        >>> for entry in segmenter(ex):
+        ...     print(entry['x'][[0, -1]])
+        [    0 31999]
+        [16000 47999]
+
+    Args:
+        length: The length of the segments in samples. If set to `-1`,
+            the original example is returned in a list of length one.
```
JanekEbb, line 64: is it possible to add a mode where length is determined automatically based on a max_length argument? E.g. if I have a signal with length 10 and I state max_length=8, then length will be set to 5 so that equally long segments are obtained?

jensheit: You could do that by setting the length and choosing one of the dictionaries from the output

tests/test_data/test_segmenter.py
```diff
@@ -0,0 +1,234 @@
...
+
+def test_copy_keys():
+    segmenter = Segmenter(length=32000, include_keys=('x', 'y'),
+                          shift=16000, copy_keys='gender')
+    ex = {'x': np.arange(65000), 'y': np.arange(65000),
+          'num_samples': 65000, 'gender': 'm'}
+    segmented = segmenter(ex)
+    assert type(segmented) == list, segmented
+    expected_keys = {key: value for key, value in ex.items()
+                     if not key == 'num_samples'}.keys()
```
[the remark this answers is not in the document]
jensheit: there is not realy a reason. I can change that.

tests/test_data/test_segmenter.py
```diff
@@ -0,0 +1,234 @@
...
+
+def test_include_to_larger():
+    segmenter = Segmenter(length=32000, shift=16000,
+                          include_keys=['x', 'y', 'z'])
+    ex = {'x': np.arange(65000), 'y': np.arange(65000),
+          'num_samples': 65000, 'gender': 'm'}
+    segmented = segmenter(ex)
+    assert type(segmented) == list, segmented
+    for idx, entry in enumerate(segmented):
+        assert all([key in entry.keys() for key in ex.keys()])
```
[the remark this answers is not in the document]
jensheit: you are right, it should throw an assert if on of the keys in include cannot be associated with any key in the input dictionary

padertorch/data/segment.py
JanekEbb: I don't think this is what I meant. What I meant is that with a max_length of 8 I, e.g., get
- for a signal of length 10: two segments of length 5
- for a signal of length 15: two segments of length 7.5
- for a signal of length 18: three segments of length 6

thequilo: I guess you mean: Maximize `length <= max_length` so that the input is fully segmented (i.e., there is nothing thrown away). This would certainly complicate the interface and I'm not sure how useful this will be in general. What's the use-case for this? If we decide in favor of this, I'd suggest adding a `length_mode` argument with possible values like "constant" and "max".

thequilo: Isn't this something that should be handled by `get_segment_boundaries` and `segment`? And if we allow this, what about spreading the segments across the input signal (i.e., determine the shift >= length automatically so that the space between segments is equal)? This is in some way a similar feature.

padertorch/data/segment.py
```diff
@@ -392,9 +424,75 @@ def get_segment_boundaries(
...
+    depending on the length_mode.
+
+    Args:
+        num_samples: number of samples to be segmented
+        length: segment length
+        shift: shift between segments
+        length_mode: constant: uses `length` for all examples
+                     max:      uses some length smaller than `length` for each
+                               example
+                     min:      uses some length larger than `length` for each
```
thequilo, line 441: Does min make sense? I'd assume that we can then just use the full length for a segment

thequilo: Or is this formulated like: maximize the number of segments with the constraint `new_length >= length`?

padertorch/data/segment.py
```diff
@@ -455,8 +555,17 @@ def moveaxis(tensor: torch.Tensor, source: int,
 
     num_samples = x.shape[axis]
     assert num_samples >= length, (num_samples, length)
-    if shift is None:
-        shift = length
+    if length_mode == 'constant':
+        end = 'cut'
+    elif length_mode in ['max', 'min']:
+        end = 'pad'
```
thequilo, line 561: I think we probably don't want to pad. In most cases, you want to ignore padded regions. Here, you don't know what is padded after calling this function.

tests/test_data/test_segmenter.py
```diff
@@ -237,3 +237,28 @@ def test_wildcard_exclude():
...
+                                    np.arange(0, new_length[idx][mode]))
+    new_length = [{'constant': 950, 'max': 947, 'min': 950},
+                  {'constant': 950, 'max': 950, 'min': 953}]
+    for mode in ['constant', 'max', 'min']:
+        for idx, ex in enumerate(examples):
+            segmenter = Segmenter(length=950, shift=250, include_keys=('x'),
+                                  length_mode=mode)
+            segmented = segmenter(ex)
+            np.testing.assert_equal(segmented[0]['x'],
+                                    np.arange(0, new_length[idx][mode]))
```
thequilo, line 264: Newline at end of file

thequilo, line 264: Can you add a test with `shift > length`?

padertorch/data/segment.py
jensheit: yes, we look at the number of times length fits into num_samples and than extent the length by the residual divided by the number of time length fits in to num_samples

padertorch/data/segment.py
jensheit: The max and min option were created for evaluation and in evaluation you should not use cut because it might change the signal.

padertorch/data/segment.py
```diff
@@ -0,0 +1,645 @@
...
+            'but is', mode
+        )
+        self.mode = mode
+        if padding:
+            # No padding is implemented for the begging of a signal
+            assert anchor in [0, 'left'], (padding, anchor)
+        self.padding = padding
+        self.flatten_separator = flatten_separator
+
+    def __call__(self, example, rng=np.random):
```
thequilo, line 155: I'm a fan of type annotations, so can you add `def __call__(self, example: dict, rng=np.random) -> List[dict]`?

thequilo: Also to the other methods?

padertorch/data/segment.py
```diff
@@ -0,0 +1,645 @@
...
+    elif isinstance(rng, callable):
+        return rng(*args, **kwargs)
+    else:
+        raise TypeError('Unknown random generator used', rng)
+
+
+def get_anchor(
+        num_samples: int, length: int, shift: int = None,
+        mode: str = 'left', rng=np.random
+):
```
thequilo, line 338: I just noticed that the return type is not annotated. Can you add that?

thequilo (commented):
I think I'm fine with the functionality now, I just noticed that some functions don't have type annotations yet.

padertorch/data/segment.py
jensheit: done

padertorch/data/segment.py
jensheit: done

PR #88, opened 2020-11-30
Improve tbx utils

thequilo:
- Add a `log` option to `spectrogram_to_image`
- Improve type annotations and docstrings
- Add `matplotlib_figure_to_image`

### DIFF
--- a/padertorch/summary/tbx_utils.py
+++ b/padertorch/summary/tbx_utils.py
@@ -1,9 +1,12 @@
 import operator
+from typing import Union, Optional, Tuple
 
 import numpy as np
 import torch
 
 from padertorch.utils import to_numpy
+
+_T_input = Union[torch.Tensor, np.ndarray]
 
 __all__ = [
     'mask_to_image',
@@ -53,14 +56,28 @@ def _apply_origin(image, origin):
     return image
 
 
-def mask_to_image(mask, batch_first=False, color=None, origin='lower'):
+def mask_to_image(
+        mask: _T_input, batch_first: bool = False, color: Optional[str] = None,
+        origin: str = 'lower'
+) -> np.ndarray:
     """
-    For more details of the output shape, see the tensorboardx docs
-    Args:
-        mask: Shape (frames, batch [optional], features)
-        batch_first: if true mask shape (batch [optional], frames, features]
+    Creates an image from a mask `Tensor` or `ndarray`.
 
-    Returns: Shape(color, features, frames)
+    For more details of the output shape, see the tensorboardx docs
+
+    Args:
+        mask: Mask to plot
+        batch_first: If `True`, `signal` is expected to have shape
+            `(batch [optional], frames, features)`. If `False`, the batch axis
+            is assumed to be in the second position, i.e.,
+            `(frames, batch [optional], features)`.
+        color: A color map name. The name is forwarded to
+               `matplotlib.pyplot.cm.get_cmap` to get the color map. If `None`,
+               grayscale is used.
+        origin: Origin of the plot. Can be `'upper'` or `'lower'`.
+
+    Returns:
+        Colorized image with shape (color (1 or 3), features, frames)
 
     """
     mask = to_numpy(mask, detach=True)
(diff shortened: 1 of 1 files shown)


padertorch/summary/tbx_utils.py
```diff
@@ -151,50 +181,73 @@ def __call__(self, image, color):
...
-        batch_first: if true mask shape (batch [optional], frames, features]
+        signal: Spectrogram to plot.
+        batch_first: If `True`, `signal` is expected to have shape
+            `(batch [optional], frames, features)`. If `False`, the batch axis
+            is assumed to be in the second position, i.e.,
+            `(frames, batch [optional], features)`.
         color: A color map name. The name is forwarded to
                `matplotlib.pyplot.cm.get_cmap` to get the color map.
+        origin: Origin of the plot. Can be `'upper'` or `'lower'`.
+        log: If `True`, the spectrogram is plotted in log domain.
```
boeddeker, line 202: Maybe add: `... log domain and shows a 50 dB range.`?

padertorch/summary/tbx_utils.py
```diff
@@ -53,14 +56,28 @@ def _apply_origin(image, origin):
     return image
 
 
-def mask_to_image(mask, batch_first=False, color=None, origin='lower'):
+def mask_to_image(
+        mask: _T_input, batch_first: bool = False, color: Optional[str] = None,
+        origin: str = 'lower'
```
boeddeker, line 61: One line per argument?

padertorch/summary/tbx_utils.py
```diff
@@ -210,14 +263,56 @@ def audio(signal, sampling_rate: int = 16000, batch_first=False,
...
-        scalars: dict=None,
-        histograms: dict=None,
-        audios: dict=None,
-        images: dict=None,
+        loss: torch.Tensor = None,
+        losses: dict = None,
+        scalars: dict = None,
+        histograms: dict = None,
+        audios: dict = None,
+        images: dict = None,
```
boeddeker, line 315: add `figures`?

thequilo: What's the difference to `images`?

padertorch/summary/tbx_utils.py
```diff
@@ -210,14 +263,56 @@ def audio(signal, sampling_rate: int = 16000, batch_first=False,
...
+        plots to reduce memory consumption!
+
+    Args:
+        figure: matplotlib figure object. If `None`, it defaults to `plt.gcf()`.
+
+    Returns:
+        Numpy array with shape (color (4), width, height)
+
+    Examples:
+        >>> from matplotlib import pyplot as plt
```
boeddeker, line 289: Add lines, how to disable GUI backend?
e.g.
`>>> # import matplotlib as mlp`
`>>> # mlp....  # Disable GUI backend, must be executed before the first "import matplotlib.pyplot"`

padertorch/summary/tbx_utils.py
```diff
@@ -210,14 +263,56 @@ def audio(signal, sampling_rate: int = 16000, batch_first=False,
     return signal, sampling_rate
 
 
+def matplotlib_figure_to_image(
```
boeddeker, line 266: Can you motivate, why `matplotlib_figure_to_image` and not `matplotlib_figure`?
i.e. tensorboradX has also a figure option. They use much more code for the conversion: https://github.com/lanpa/tensorboardX/blob/8f50866d10ca064b7093c364985aa25d1dcd27a9/tensorboardX/utils.py#L2

padertorch/summary/tbx_utils.py
boeddeker: It supports matplotlib figures

padertorch/summary/tbx_utils.py
thequilo: I didn't know that function. It could work equally well and additionally supports multiple figures and doesn't require setting the backend before calling this function

padertorch/summary/tbx_utils.py
thequilo: Hmm, now that I found this option, I'm not sure if the new function is required anymore. I don't see a strong reason to draw the figure in the review function rather than drawing it in the hook. If the user wants to draw the figure in review, the user can still use the `figure_to_image` function from tensorboardX.

It would be nice to document this somewhere.

padertorch/summary/tbx_utils.py
thequilo: Done

padertorch/summary/tbx_utils.py
thequilo: Added `figures`

padertorch/summary/tbx_utils.py
```diff
@@ -210,42 +267,134 @@ def audio(signal, sampling_rate: int = 16000, batch_first=False,
...
+        (3, 480, 640)
+    """
+    from tensorboardX.utils import figure_to_image as tbX_figure_to_image
+    return tbX_figure_to_image(figure(fig, close=close), close=False)
+
+
+def figure(
+        fig: 'matplotlib.figure.Figure' = None,
+        close=True,
+) -> 'matplotlib.figure.Figure':
```
thequilo, line 319: I'm not sure if these type annotations make sense here. Matplotlib is not imported to keep the import overhead small, but then these annotations are not parsed correctly by PyCharm.

boeddeker: Does PyCharm complain, that this is an error? Or is it just a warning?
Maybe `# noqa` can suppress pycharm's problmes.

I think the type annotation for this function is not so important. It should be obvious, what the type is.

boeddeker (approved):

padertorch/summary/tbx_utils.py
thequilo: Not required anymore

padertorch/summary/tbx_utils.py
thequilo: It's just a warning and if you use the function, it doesn't infer the return type correctly. But that's just a minor issue.

PR #89, opened 2020-12-10
Trainer: Robustify log_error_state for more pytorch versions and add test

boeddeker:
@TCord reported that his pytorch used the legacy save from pytorch (Unknown, why).
This PR fixes `log_error_state` from the `Trainer` to work with the legacy save.

This PR also adds tests for this.

### DIFF
--- a/padertorch/train/trainer.py
+++ b/padertorch/train/trainer.py
@@ -624,7 +624,7 @@ class Trainer(Configurable):
 
         return loss, review
 
-    def log_error_state(self, data_dict, folder='log', stdout=sys.stdout):
+    def log_error_state(self, data_dict, folder='log', file=sys.stdout):
         """
 
         Args:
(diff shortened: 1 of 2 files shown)


padertorch/train/trainer.py
```diff
@@ -608,7 +624,7 @@ def _review_to_loss_and_summary(self, review):
 
         return loss, review
 
-    def log_error_state(self, data_dict, folder='log'):
+    def log_error_state(self, data_dict, folder='log', stdout=sys.stdout):
```
thequilo, line 627: I'd call this `file` (as in `print`) or `stream` or similar. `stdout` gets confusing if you set it to `stderr`.

boeddeker: Good point. I changed it. I only introduced it to avoid capturing the stdout in the test. Since you now mention it, should the default be `stderr`? It is used to report problems in the failure logging progress.

jensheit (approved):

padertorch/train/trainer.py
thequilo: It's a little bit late now, but probably yes. The python `logging` module also uses `sys.stderr` as the default output stream.

boeddeker: Yes, 1 second after I merged, I remembered this open point.
Ok, I will change it in the future.

PR #91, opened 2020-12-18
Tasnet Segmenter

thequilo:
Uses the Segmenter in the TasNet example. I never used the local shuffle buffer before. Is the buffer size of 128 reasonable?

### DIFF
--- a/padertorch/contrib/examples/source_separation/tasnet/train.py
+++ b/padertorch/contrib/examples/source_separation/tasnet/train.py
@@ -210,9 +210,9 @@ def pre_batch_transform(inputs):
     }
 
 
-def prepare_iterable(
-        db, dataset: str, batch_size, chunk_size, prefetch=True,
-        dataset_slice=None
+def prepare_dataset(
+        db, dataset: str, batch_size, chunk_size, shuffle=True,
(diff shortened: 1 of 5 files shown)


boeddeker (approved):

padertorch/contrib/examples/source_separation/tasnet/train.py
```diff
@@ -316,8 +331,8 @@ def prepare_and_train(_run, _log, trainer, train_dataset, validate_dataset,
 
     # Perform a test run to check if everything works
     trainer.test_run(
-        prepare_iterable_captured(db, train_dataset),
-        prepare_iterable_captured(db, validate_dataset),
+        prepare_dataset_captured(db, train_dataset, shuffle=True),
+        prepare_dataset_captured(db, validate_dataset, shuffle=False),
```
jensheit, line 335: I was thinking whether one should deactivate the segmenter during validation

jensheit (approved):

padertorch/contrib/examples/source_separation/tasnet/train.py
thequilo: It's now deactivated during validation.

PR #92, opened 2021-01-06
Allow integer as sequence_lenghts in compute_masks

jensheit:
adds possiblity to use integers as sequence_lengths in compute_masks

### DIFF
--- a/padertorch/ops/sequence/mask.py
+++ b/padertorch/ops/sequence/mask.py
@@ -1,13 +1,17 @@
 import torch
+from typing import Union, List
 
 
-def compute_mask(x, sequence_lengths, batch_axis=0, sequence_axis=1):
+def compute_mask(x, sequence_lengths: Union[List[int], int] = None,
+                 batch_axis=0, sequence_axis=1):
     """
-    This function calculates a mask which indicates the position of non-padded values.
+    This function calculates a mask which indicates the position
+    of non-padded values.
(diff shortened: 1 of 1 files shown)


padertorch/ops/sequence/mask.py
```diff
@@ -1,13 +1,18 @@
 import torch
+from typing import Union, List
 
 
-def compute_mask(x, sequence_lengths, batch_axis=0, sequence_axis=1):
+def compute_mask(x: torch.Tensor,
+                 sequence_lengths: Union[List[int], int, None],
+                 batch_axis: int = 0, sequence_axis: int = 1):
```
boeddeker, line 7: Can you add a line break between `batch_axis` and `sequence_axis`?
```python
def compute_mask(
        x: torch.Tensor,
        sequence_lengths: Union[List[int], int, None],
        batch_axis: int = 0,
        sequence_axis: int = 1,
):
```

boeddeker:
What is the motivation of `sequence_lenghts: int`?
I expected that `sequence_lenghts is None` uses the faster/better code path, while yielding the same result.

jensheit:
> What is the motivation of `sequence_lenghts: int`?
> I expected that `sequence_lenghts is None` uses the faster/better code path, while yielding the same result.

In theory there might be some array I have padded for convolution or other reasons, where I all original signals have the same length.

PR #93, opened 2021-01-12
tbx_utils: Sync spectrogram_to_image with paderbox and made visible_dB an option

boeddeker:
- `spectrogram_to_image` now expects power as input (same as in `paderbox`)
 - Made `visible_dB` an option
 - Add comparison with `paderbox.visualization.plot.stft`

### DIFF
--- a/padertorch/summary/tbx_utils.py
+++ b/padertorch/summary/tbx_utils.py
@@ -119,7 +119,7 @@ def stft_to_image(
         Colorized image with shape (color (1 or 3), features, frames)
 
 
-    Comparison with `paderbox.visualization.plot.stft`
+    Small test to see the effect of `visible_dB`:
 
         >>> visible_dB = 60
         >>> 10 ** (-visible_dB / 20)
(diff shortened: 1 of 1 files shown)


padertorch/summary/tbx_utils.py
```diff
@@ -109,14 +110,38 @@ def stft_to_image(
...
+        >>> np.squeeze(stft_to_image(
+        ...     np.array(data)[:, None], color=None, visible_dB=60))
+        array([255,  51,  40,   1,   0], dtype=uint8)
+
     """
     signal = to_numpy(signal, detach=True)
 
     return spectrogram_to_image(
-        np.abs(signal), batch_first=batch_first, color=color, origin=origin,
+        signal.real ** 2 + signal.imag ** 2,
```
thequilo, line 140: What's the difference between `np.abs(signal)` and `signal.real ** 2 + signal.imag ** 2`? I would expect that `np.abs` is the better option.

boeddeker: I made two changes here:
 - I changed from `np.abs(signal)` to `np.abs(signal)**2`, because we usually define the spectrogram as the power of the STFT.
 - I changed `np.abs(signal)**2` to `signal.real ** 2 + signal.imag ** 2`:
    - Use the same operation as in paderbox
    - `np.abs` includes a square root operation, this avoids such an operation. If I remember correct, that operation is faster for common shapes that we have.

thequilo: Ok, I didn't notice the missing square root. Then, it makes sense.

thequilo (approved):

PR #95, opened 2021-01-14
fix examples

JanekEbb:


padertorch/contrib/examples/audio_synthesis/wavenet/evaluate.py
```diff
@@ -70,10 +69,11 @@ def main(_run, exp_dir, storage_dir, database_json, test_set, max_examples, devi
             x = model.feature_extraction(example['stft'], example['seq_len'])
             x = model.wavenet.infer(
                 x.squeeze(1),
-                chunk_length=80_000,
+                chunk_length=48_000,
                 chunk_overlap=16_000,
             )
-            assert target.shape == x.shape, (target.shape, x.shape)
+            assert config['stft']['shift'] > (x.shape[-1] - target.shape[-1]) >= 0, (target.shape, x.shape)
+            x = x[..., :target.shape[-1]]
```
michael-kuhlmann, line 76: Can you add a comment on why this assertion and cutting of `x` is necessary? I think it's because the WaveNet can only generate sequence lengths that are multiples of `shift`?

michael-kuhlmann (approved):

PR #101, opened 2021-03-22
fixes missing activation in covnet

jensheit:


thequilo:
Can you also fix the link in the named config `convnet` in `padertorch.contrib.examples.tasnet.source_separation.train`?

thequilo:
Is this activation correct? In the Conv-TasNet paper, there is no activation after the temporal convolutional network, but after the output projection that is located in the TasNet model. Also, `Conv1d` already contains an activation function, so this activation functions is basically applied twice (`relu(relu(last_block_out) + last_block_in)`).

TCord:
If I understood the code correctly, this `activation_fn` can be chosen differently from the one in the `Conv1d` layer. With default settings, it is redundant, because a `relu` is applied twice, but e.g. when employing the `Convnet` with STFT/iSTST encoder/decoder it might be sensible to choose a sigmoid as activation and make the masks in the latent space better interpretable.

jensheit:
You are both right it is a little bit redundant. There a two solutions, one we just remove the option and say that this would be a task for the model and not for the module or we remove the activation from the last cnn block.

thequilo:
> If I understood the code correctly, this `activation_fn` can be chosen differently from the one in the `Conv1d` layer. With default settings, it is redundant, because a `relu` is applied twice, but e.g. when employing the `Convnet` with STFT/iSTST encoder/decoder it might be sensible to choose a sigmoid as activation and make the masks in the latent space better interpretable.

In the `TasNet` model, we have another activation function (`convnet -> prelu -> 1x1 conv -> activation_fn -> multiplication with encoded mixture`, as it was done in the Conv-TasNet paper) that should be set to a sigmoid in that case

thequilo:
Maybe the definition of the convnet is not super clear here, and it is not obvious if the convent should include these 1x1 convolutions or not. We chose to put it into the TasNet model so that we can replace the convnet with other architectures like DPRNN.

jensheit:
I have removed the activation, the upside to this is that already trained models still work with this change if the activation is removed from the config.json or during initialization.

thequilo (approved):

PR #102, opened 2021-05-06
Add torch_dataclass

thequilo:


padertorch/summary/tbx_utils.py
```diff
@@ -84,6 +88,13 @@ def mask_to_image(
     """
     mask = to_numpy(mask, detach=True)
 
+    if np.any(mask < 0) or np.any(mask > 1):
```
boeddeker, line 91: Can you use sum instead of any and report the number of clipped values?
I think sum should have a similar execution time.

PR #103, opened 2021-06-02
Update configurable to use partial

jensheit:
Partial is added as a special key to configurable to allow the use of non-initialized classes or functions as input to a class

### DIFF
--- a/padertorch/configurable.py
+++ b/padertorch/configurable.py
@@ -702,7 +702,8 @@ def import_class(name: [str, callable]):
 
     """
     if not isinstance(name, str):
-        assert callable(name), name
+        assert callable(name), (
+            'expects string or callcable but got', type(name), name)
         return name
 
     if '.' not in name:
--- a/padertorch/configurable.py
+++ b/padertorch/configurable.py
@@ -18,20 +18,16 @@ used for that instance in your modified `finalize_docmatic_config`.
 import sys
 import os
 import collections
+import functools
 import importlib
 import inspect
 from pathlib import Path
-from functools import partial
 import copy
 
 import paderbox as pb
 
 # pylint: disable=import-outside-toplevel
 
-# These special keys are used in the config to indicate classes or functions
-# 'factory' is used to specify initialized classes or function outputs as an
-# input. 'partial' is used if an input is a non-initialized class or
-# a functions
 
 class Configurable:
     """Allow subclasses to be configured automatically from JSON config files.
@@ -198,10 +194,13 @@ class Configurable:
     NN input size depends on selected input features).
 
     Some modules need unitilized classes or functions as an input.
+    Factory already supports functions but cannot make them json
+    serializable.
     Therefore, one can use the partial key instead of the factory key.
     All defined kwargs will overwrite the defaults without calling the function
     or initializing the class using partial.
-    One example is SpeechBrain which requires the activity to be not
+    This is essential a functions.partial call.
+    One usecase is SpeechBrain which requires the activity to be not
     initialized at the class input.
 
         >>> class SBDenseLayer(Configurable, torch.nn.Module):
@@ -927,7 +926,7 @@ def recursive_class_to_str(config, sort=False):
     # ToDo: Support tuple and list?
     if isinstance(config, dict):
         d = config.__class__()
-        special_key = _get_special_key(config.keys())
+        special_key = _get_special_key(config)
 
         if sort and special_key:
             # Force factory to be the first key
--- a/padertorch/configurable.py
+++ b/padertorch/configurable.py
@@ -573,8 +573,7 @@ def fix_doctext_import_class(locals_dict):
         ImportError: Could not import 'Foo' from '...configurable',
         because module '...configurable' has no attribute 'Foo'
         <BLANKLINE>
-        Make su
-        re that
+        Make sure that
          1. This is the class you want to import.
          2. You activated the right environment.
          3. The module exists and has been installed with pip.
--- a/padertorch/configurable.py
+++ b/padertorch/configurable.py
@@ -199,7 +199,7 @@ class Configurable:
     Therefore, one can use the partial key instead of the factory key.
     All defined kwargs will overwrite the defaults without calling the function
     or initializing the class using partial.
-    This is essential a functions.partial call.
+    This is essentially a `functools.partial` call.
     One usecase is SpeechBrain which requires the activity to be not
     initialized at the class input.
 
@@ -702,8 +702,9 @@ def import_class(name: [str, callable]):
 
     """
     if not isinstance(name, str):
-        assert callable(name), (
-            'expects string or callcable but got', type(name), name)
+        if not callable(name):
+            raise TypeError(
+                'expects string or callcable but got', type(name), name)
         return name
 
     if '.' not in name:
@@ -891,12 +892,12 @@ def class_to_str(cls, fix_module=False):
 
 def recursive_class_to_str(config, sort=False):
     """
-    Ensures that all factory values are strings.
+    Ensures that factory and partial values are strings.
 
     The config that is returned from a configurable already takes care, that
-    all factory values are strings. But when sacred overwrites a factory values
-    with a class and not the str, the config will contain a class instead of
-    the corresponding string.
+    all factory and partial values are strings. But when sacred overwrites a
+    factory values with a class and not the str, the config will contain a
+    class instead of the corresponding string.
 
     Args:
         config:
@@ -930,10 +931,11 @@ def recursive_class_to_str(config, sort=False):
         special_key = _get_special_key(config)
 
         if sort and special_key:
-            # Force factory to be the first key
+            # Force the special key to be the first key
             d[special_key] = None  # will be set later
             imported = import_class(config[special_key])
             arg_names = inspect.signature(imported).parameters.keys()
+            # This ensure that the keys are in the same order as the signature
             for k in arg_names:
                 if k in config:
                     d[k] = None  # will be set later
@@ -959,8 +961,8 @@ def _split_factory_kwargs(config, key='factory'):
 
 
 def _get_special_key(config):
-    # These special keys are used in the config to indicate classes or
-    # functions 'factory' is used to specify initialized classes or
+    # These special keys are used in the config to indicate a class or
+    # function. 'factory' is used to specify initialized classes or
     # function outputs as an input. 'partial' is used if an input is a
     # non-initialized class or a functions
     for key in ['factory', 'partial']:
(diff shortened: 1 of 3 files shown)


padertorch/configurable.py
```diff
@@ -325,15 +369,18 @@ def get_config(
...
             config['factory'] = import_class(config['factory'])
             if inspect.isclass(config['factory']) \
                     and issubclass(config['factory'], Configurable):
                 # When subclass of Configurable expect proper subclass
-                assert issubclass(import_class(config['factory']), cls), (
+                assert issubclass(config['factory'], cls), (
                     config['factory'], cls)
+        elif 'partial' in config:
+            config['partial'] = import_class(config['partial'])
+            assert callable(config['partial']), config['partial']
```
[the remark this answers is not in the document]
jensheit: Not if we want to allow functions here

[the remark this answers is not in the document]
thequilo: Functions are allowed in `factory` as well, so that's not a problem. The check checks that if `factory` is `Configurable`, then it should be a sub-class of `cls`, so that `A.from_config(...)` always returns an instance of a sub-class of `A` if `factory` is a `Configurable`.

padertorch/configurable.py
```diff
@@ -960,6 +1024,17 @@ def config_to_instance(config, strict=False):
 
     >>> import torch.nn
     >>> config = {
+    ...     'partial': 'torch.nn.modules.activation.ReLU',
```
boeddeker, line 1027: Reorder: `partial` should be the last example.

padertorch/configurable.py
```diff
@@ -998,19 +1073,30 @@ def config_to_instance(config, strict=False):
...
+                factory = import_class(factory)
+                kwargs = config_to_instance(kwargs, strict)
+
+                _check_factory_signature_and_kwargs(factory, kwargs, strict)
+
+                if key == 'factory':
+                    new = factory(**kwargs)
+                elif key == 'partial':
+                    if len(kwargs) > 0:
+                        from functools import partial
```
boeddeker, line 1088: Use a global import of `functools`

padertorch/configurable.py
```diff
@@ -998,19 +1073,30 @@ def config_to_instance(config, strict=False):
...
+                        new = partial(factory, **kwargs)
+                    else:
+                        new = factory
```
boeddeker, line 1091: Raise `Exception('Can not happen.')`

padertorch/configurable.py
```diff
@@ -21,12 +21,18 @@
...
 import copy
 
 import paderbox as pb
 
 # pylint: disable=import-outside-toplevel
 
+# These special keys are used in the config to indicate classes or functions
+# 'factory' is used to specify initialized classes or function outputs as an
+# input. 'partial' is used if an input is a non-initialized class or
+# a functions
```
thequilo, line 34: Why this comment here?

padertorch/configurable.py
```diff
@@ -892,12 +952,21 @@ def recursive_class_to_str(config, sort=False):
...
+def _split_factory_kwargs(config, key='factory'):
     kwargs = config.copy()
-    factory = kwargs.pop('factory')
+    factory = kwargs.pop(key)
     return factory, kwargs
 
 
+def _get_special_key(config):
+    for key in ['factory', 'partial']:
+        if callable(config):
```
thequilo, line 963: Why do you need `callable` here? This calls `config` two times.

padertorch/configurable.py
```diff
@@ -21,12 +21,18 @@
 import importlib
 import inspect
 from pathlib import Path
+from functools import partial
```
boeddeker, line 24: How about using `import functools` to avoid confusion with the keyword `partial`?

padertorch/configurable.py
```diff
@@ -21,12 +21,18 @@
...
 import copy
 
 import paderbox as pb
 
 # pylint: disable=import-outside-toplevel
 
+# These special keys are used in the config to indicate classes or functions
```
boeddeker, line 31: ?
Have you forgot to delete this comment?

padertorch/configurable.py
```diff
@@ -191,6 +197,44 @@ class Configurable:
     Another use case for this behaviour are depended config entries (e.g.
     NN input size depends on selected input features).
 
+    Some modules need unitilized classes or functions as an input.
```
boeddeker, line 200: Say that it is essentially a `functools.partial` call?

padertorch/configurable.py
```diff
@@ -1317,13 +1423,13 @@ def get_sub_config(self, key, mutable_idx=None):
         else:
             raise KeyError(key)
 
-    def _key_candidates(self):
-        if 'factory' in self.data:
-            factory = import_class(self.data['factory'])
+    def _key_candidates(self, special_key='factory'):
+        if special_key in self.data:
```
thequilo, line 1427: Can't you use `_get_special_key` here?

padertorch/configurable.py
```diff
@@ -191,6 +197,44 @@ class Configurable:
     Another use case for this behaviour are depended config entries (e.g.
     NN input size depends on selected input features).
 
+    Some modules need unitilized classes or functions as an input.
+    Therefore, one can use the partial key instead of the factory key.
+    All defined kwargs will overwrite the defaults without calling the function
+    or initializing the class using partial.
+    One example is SpeechBrain which requires the activity to be not
+    initialized at the class input.
```
boeddeker, line 205: Mention, that we support functions as arguments, but to support json/yaml serialization, we recommend the partial?

padertorch/configurable.py
```diff
@@ -1510,19 +1618,21 @@ def get(self, key, default=None):
     def to_dict(self):
         """Export the Configurable object to a dict."""
         result_dict = {}
-        if 'factory' in self._key_candidates():
-            self._update_factory_kwargs()
-        for k in self._key_candidates():
+        special_key = _get_special_key(self._key_candidates)
```
thequilo, line 1621: Why this? Isn't `_get_special_key(self.data.keys())` enough?

padertorch/configurable.py
```diff
@@ -892,12 +951,22 @@ def recursive_class_to_str(config, sort=False):
...
 
 
+def _get_special_key(config):
+    # These special keys are used in the config to indicate classes or
+    # functions 'factory' is used to specify initialized classes or
+    # function outputs as an input. 'partial' is used if an input is a
+    # non-initialized class or a functions
+    for key in ['factory', 'partial']:
+        if key in config.keys():
+            return key
```
boeddeker, line 967: Add a verbose `return None`? i.e. highlight that the return value is one of: `'factory'`, `'partial'`, `None`

padertorch/configurable.py
```diff
@@ -1348,9 +1456,9 @@ def keys(self):
         return tuple(self.data.keys())
 
     def _check_redundant_keys(self, msg):
-        assert 'factory' in self.data
-        factory = import_class(self.data['factory'])
-        parameters = inspect.signature(factory).parameters.values()
+        assert self.special_key in self.data, self.data
```
boeddeker, line 1459: `None in {None: ...}` returns `True`.
While this is unlikely, `assert self.special_key is not None, f'Missing factory or partial in {self.data}'` does the correct check.

boeddeker: Or `assert self.special_key, f'Missing factory or partial in {self.data}'` as you used at other positions.

padertorch/configurable.py
```diff
@@ -1411,10 +1519,10 @@ def setdefault(self, key, default):
         return self[key]
 
     def _update_factory_kwargs(self):
-        assert 'factory' in self.data, self.data
+        assert self.special_key in self.data, self.data
```
boeddeker, line 1522: Same as above. Not the Exception without an explicit `str` is not informative.

padertorch/configurable.py
```diff
@@ -191,6 +193,47 @@ class Configurable:
     Another use case for this behaviour are depended config entries (e.g.
     NN input size depends on selected input features).
 
+    Some modules need unitilized classes or functions as an input.
+    Factory already supports functions but cannot make them json
+    serializable.
+    Therefore, one can use the partial key instead of the factory key.
+    All defined kwargs will overwrite the defaults without calling the function
+    or initializing the class using partial.
+    This is essential a functions.partial call.
```
thequilo, line 202: `functools.partial` (with backticks and functools instead of functions)

thequilo: And "essentially"

padertorch/configurable.py
```diff
@@ -654,7 +702,8 @@ def import_class(name: [str, callable]):
 
     """
     if not isinstance(name, str):
-        assert callable(name), name
+        assert callable(name), (
+            'expects string or callcable but got', type(name), name)
```
thequilo, line 706: This shouldn't be `assert` but raise a `TypeError` because it checks user input

padertorch/configurable.py
```diff
@@ -864,22 +913,33 @@ def recursive_class_to_str(config, sort=False):
...
         d = config.__class__()
-        if sort and 'factory' in config:
+        special_key = _get_special_key(config)
+
+        if sort and special_key:
             # Force factory to be the first key
-            d['factory'] = None  # will be set later
-            factory = import_class(config['factory'])
-            arg_names = inspect.signature(factory).parameters.keys()
+            d[special_key] = None  # will be set later
```
thequilo, line 934: The comment above this line still uses "factory", should be "special key"

thequilo: The following code is a little hard to understand. Maybe little longer comment that this is only done to ensure that the ordering of the entries equals the ordering of the signature could help

padertorch/configurable.py
```diff
@@ -892,12 +952,23 @@ def recursive_class_to_str(config, sort=False):
...
+def _split_factory_kwargs(config, key='factory'):
     kwargs = config.copy()
-    factory = kwargs.pop('factory')
+    factory = kwargs.pop(key)
     return factory, kwargs
 
 
+def _get_special_key(config):
+    # These special keys are used in the config to indicate classes or
+    # functions 'factory' is used to specify initialized classes or
```
thequilo, line 963: Is there a fullstop missing?

padertorch/configurable.py
```diff
@@ -892,12 +952,23 @@ def recursive_class_to_str(config, sort=False):
...
+    # function outputs as an input. 'partial' is used if an input is a
+    # non-initialized class or a functions
```
thequilo, line 965: "functions" -> "function"

padertorch/configurable.py
```diff
@@ -995,17 +1066,44 @@ def config_to_instance(config, strict=False):
...
+        if special_key:
+            factory, kwargs = _split_factory_kwargs(config, key=special_key)
+            if isinstance(factory, str):
+                factory = import_class(factory)
+            elif callable(factory):
+                pass
+            else:
+                raise TypeError(f'The special key {special_key} expects a'
+                                f' string or a callable but got',
+                                type(factory), factory)
```
thequilo, line 1093: You could here also do (after changing the assertion to a `TypeError` in `import_class`)
```python
try:
    factory = import_class(factory)
except TypeError as e:
    raise TypeError(...) from e
```
It is a little bit confusing that this check is made twice. But, the `try`/`except` variant might create confusing errors if a `TypeError` is raised inside `import_class` that is not caused by the type of factory.

jensheit: I am partial to both solutions. At the moment I do not see a possibility for a TypeError inside of import_class that is not caused by the factory type. So I will change this to `try / except`

thequilo (approved):

boeddeker (approved):

PR #106, opened 2021-06-22
Allow sequence lengths of type tensor

alexanderwerning:
Specifying sequence lengths for normalization with padertorch.modules.normalization.Normalization in a torch.tensor instead of a List raises an error currently. Both Lists and Tensors should be acceptable.

### DIFF
--- a/padertorch/ops/sequence/mask.py
+++ b/padertorch/ops/sequence/mask.py
@@ -62,8 +62,7 @@ def compute_mask(x, sequence_lengths, batch_axis=0, sequence_axis=1):
         sequence_axis = x.dim() + sequence_axis
     if not torch.is_tensor(sequence_lengths):
         sequence_lengths = torch.Tensor(sequence_lengths).long().to(x.device)
-    elif sequence_lengths.device != x.device:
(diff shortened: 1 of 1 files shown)


padertorch/ops/sequence/mask.py
```diff
@@ -60,10 +60,13 @@ def compute_mask(x, sequence_lengths, batch_axis=0, sequence_axis=1):
         batch_axis = x.dim() + batch_axis
     if sequence_axis < 0:
         sequence_axis = x.dim() + sequence_axis
-    sequence_lengths = torch.Tensor(sequence_lengths).long().to(x.device)
+    if not torch.is_tensor(sequence_lengths):
+        sequence_lengths = torch.Tensor(sequence_lengths).long().to(x.device)
+    elif sequence_lengths.device != x.device:
```
jensheit, line 65: I think we do not need this elif branch because this should only happen if there is a prior error in the code which might be covered up by this. You can add an assert, that the tensors are on the same device, but I do not feel it is necessary.

jensheit (approved):
after removing the elif branch the change is fine

PR #107, opened 2021-06-25
Allow partial if not all parameters are set

jensheit:
Updates Configurable to allow partial to work if not all parameters are set


boeddeker:
Can you use `inspect.Signature.bind_partial` for `partial` instead of deactivating the test?

jensheit:
> Can you use `inspect.Signature.bind_partial` for `partial` instead of deactivating the test?
good idea, it's added

boeddeker (approved):

PR #118, opened 2021-09-27
fix example_to_device

boeddeker:


boeddeker:
Ups, the commit name should be example_to_numpy

PR #119, opened 2021-09-28
Add sa-SDR loss function

thequilo:


padertorch/ops/losses/regression.py
```diff
@@ -277,5 +277,21 @@ def log1p_mse_loss(estimate: torch.Tensor, target: torch.Tensor,
...
 
-# TODO: Add log1p_mse loss from interspeech paper
+
+def source_aggregated_sdr_loss(
+        estimate: torch.Tensor,
+        target: torch.Tensor,
+) -> torch.Tensor:
+    """
+    The source-aggregated SDR loss. There is no `reduction` argument because the loss always
+    combines the losses for multiple paris of targets and estimates.
```
boeddeker, line 287: `paris` -> `pairs`

padertorch/ops/losses/regression.py
```diff
@@ -277,5 +277,21 @@ def log1p_mse_loss(estimate: torch.Tensor, target: torch.Tensor,
...
     )
 
-# TODO: Add log1p_mse loss from interspeech paper
+
+def source_aggregated_sdr_loss(
+        estimate: torch.Tensor,
+        target: torch.Tensor,
+) -> torch.Tensor:
+    """
+    The source-aggregated SDR loss. There is no `reduction` argument because the loss always
```
boeddeker, line 286: Maybe you can highlight, that the loss is equal for reduction `mean` and `sum`?

padertorch/ops/losses/regression.py
```diff
@@ -277,5 +277,21 @@ def log1p_mse_loss(estimate: torch.Tensor, target: torch.Tensor,
...
+    combines the losses for multiple paris of targets and estimates.
+    """
+    # Calculate the source-aggregated SDR: Sum the squares of all targets and
+    # all errors before computing the ratio.
+    sa_sdr = 10 * torch.log10(
+        torch.sum(target**2) / torch.sum((estimate - target)**2)
```
boeddeker, line 292: Question: Is mean more stable?

thequilo: I don't think that it makes a difference. I just tested with very large input shapes and input values and couldn't observe a difference larger than 1e-6.

boeddeker: ok, great. I am not sure, if an overflow may be a problem for float32, but I also don't know if the mean is overflow stable.

boeddeker (approved):

PR #133, opened 2022-02-11
Windows tests

alexanderwerning:
Fixes problems on windows, mainly path issues ('/' vs '\'). All training tests are skipped since symlinks are not supported on Windows (without Administrator privileges). Some depreciation warnings were also fixed.


.github/workflows/windows-tests.yml
```diff
@@ -0,0 +1,42 @@
...
+      run: |
+        # stop the build if there are Python syntax error or undefined names
+        #flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+        # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
+        #flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
+    - name: Test with pytest
+      run: |
+        pytest -v "tests/" "padertorch/"
+        python -m coverage xml --include="padertorch*"
+    - name: Codecov
```
boeddeker, line 40: This can be removed. Coverage should be reported from the linux tests.

boeddeker (approved):

tests/test_train/test_hooks.py
```diff
@@ -86,6 +87,9 @@ def train_loop_epoch(self, length, max_it_len):
 
 
 def test_summary_hook():
+    if sys.platform.startswith('win'):
+        pytest.skip('this doctest does not work on Windows, '
+                    'training is not possible on Windows due to symlinks being unavailable')
```
thequilo, line 92: Can you explain why this is the case? E.g., add a comment or change the skip message to something like "Training is not possible on Windows because file symlinks are unavailable for non-admin users" (I believe that this is the reason, not 100% sure)

boeddeker: You can add a new function below the `padertorch.testing` namespace and add there the comments. In this way, the skip message can stay short, and the source code has more details.

PR #135, opened 2022-05-04
add freeze method in Conv and Normalization Modules

JanekEbb:
(and use clamping in normalization.py to prevent division by zero)


jensheit (approved):
I have concentrated on the changes in normalization, since the other changes are in contrib

PR #136, opened 2022-06-24
add optional codebarbon emissions tracking to trainer

JanekEbb:


padertorch/train/hooks.py
```diff
@@ -1022,3 +1023,46 @@ def set_value(self, trainer, value):
...
+
+    def pre_step(self, trainer: 'pt.Trainer'):
+        if self.trigger(iteration=trainer.iteration, epoch=trainer.epoch):
+            self.tracker.flush()
+            self.dump_emissions(trainer)
+
+    def dump_emissions(self, trainer):
+        emissions_file = Path(self.storage_dir) / 'emissions.csv'
+        with emissions_file.open() as fid:
+            lines = fid.read().split('\n')
```
thequilo, line 1050: Is it possible to read the data from the tracker object directly? If so, I would prefer to avoid a file access

padertorch/train/hooks.py
```diff
@@ -1022,3 +1023,46 @@ def set_value(self, trainer, value):
...
+        if len(lines) < 3:
+            return
+        lines = lines[:-1]
```
thequilo, line 1053: How often does the tracker update the file? Does this catch all updates?

padertorch/train/hooks.py
```diff
@@ -1022,3 +1023,46 @@ def set_value(self, trainer, value):
...
+            key: value for key, value in zip(
+                lines[0].split(','), lines[-1].split(','))
+        }
+        for key in [
+            'emissions', 'cpu_energy', 'gpu_energy', 'ram_energy',
+            'energy_consumed'
+        ]:
+            tag = f'{self.prefix}/{key}'
+            trainer.writer.add_scalar(
+                tag, float(emissions[key]), trainer.iteration)
```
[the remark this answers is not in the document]
JanekEbb: Not sure if that would work as expected, as the summary is dumped in the pre_step

padertorch/train/hooks.py
JanekEbb: Yea I would have preferred so too, but this is only possible when calling private methods (_measure_power_and_energy and _prepare_emissions_data) when I understand correctly.

JanekEbb: unfortunately, the flush() method only returns emissions_data.emissions (which are the CO2 emissions in kgs) rather than all the emissions_data including gpu_energy, cpu_energy and ram_energy

padertorch/train/hooks.py
JanekEbb: it is updated when the flush or stop method is called, which the hook does right before calling dump_emissions

padertorch/train/hooks.py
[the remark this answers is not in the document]
thequilo: It is also dumped in `close`, so it should work

padertorch/train/hooks.py
thequilo: I would then stick to the official public interface

TCord (approved):

PR #143, opened 2022-12-22
fix ignored padding argument when fading is disabled

boeddeker:
Reported by @JanekEbb


JanekEbb:
Can you make it ```if pad and ((time_signal.shape[axis] + shift - window_length) % shift) != 0:``` to prevent padding a whole shift when ```((time_signal.shape[axis] + shift - window_length) % shift) == 0```?

PR #147, opened 2023-07-20
Fix GitHub actions

michael-kuhlmann:
GitHub actions currently do not pass because pb_bss cannot be successfully installed due to an old version of setuptools that is used. This pull request fixes this and upgrades the action versions of checkout and setup-python.


michael-kuhlmann:
If you do not like to have the contrib commits in this PR, I can first merge the contrib changes and then create a new clean PR for this issue.

boeddeker:
> If you do not like to have the contrib commits in this PR, I can first merge the contrib changes and then create a new clean PR for this issue.

We can use one PR to merge both. It is unlikely, that we have second PR, that should be merged before your `praat-textgrids` code.

.github/workflows/tests.yml
```diff
@@ -15,17 +15,17 @@ jobs:
         python-version: [3.7, 3.8]
```
boeddeker, line 15: Could you add 3.9 and 3.10? It's some time ago, since we updated the python version and in the meantime numpy dropped the support for 3.7 and 3.8 (https://numpy.org/neps/nep-0029-deprecation_policy.html).

We can keep 3.7 and 3.8, since they are still working.

michael-kuhlmann:
I have added python 3.9 and 3.10 and adapted two doctests to work with the latest versions of torch and python:

- [_ChunkRNN](https://github.com/fgnt/padertorch/blob/master/padertorch/modules/dual_path_rnn.py#L284): According to @thequilo, LSTMs in the latest torch version (tested with `torch 2.0.1`) behave non-deterministically when using packed vs. non-packed input. He suggested to fix the doctest by replacing the test for strict equality with an `allclose` test. I increased the `atol` argument to 1e-6 as the max observed absolute difference between the outputs was around 6e-7
- [dataclass_to_config](https://github.com/fgnt/padertorch/blob/8e1700ff461b5ad3d258f8f2d1d3e155139a312b/padertorch/configurable.py#L678): Python 3.10 introduced the new `kw_only` attribute for dataclasses which appears in the expected error message for Python 3.10 but not for Python<=3.9. Since we are only testing for the occurrence of the RunTimeError, I have added an ellipsis where the `kw_only` attribute would appear.

boeddeker (approved):

boeddeker:
LGTM. The azure tests can be ignored, they got somehow canceled.

PR #153, opened 2023-12-30
multiple updates

JanekEbb:


TCord (approved):

padertorch/train/hooks.py
```diff
@@ -1017,12 +1031,21 @@ def get_optimizer(self, trainer):
         return optimizer.optimizer
 
     def get_value(self, trainer):
-        return self.get_optimizer(trainer).param_groups[0]['lr']
+        opt = self.get_optimizer(trainer)
+        lrs = [param_group['lr'] for param_group in opt.param_groups]
+        if len(set(lrs)) == 1:
```
boeddeker, line 1036: Why returning a scalar, when all lrs are the same?
Why not simply `if len(lrs) == 1:`?

padertorch/train/optimizer.py
```diff
@@ -37,9 +38,11 @@ def clip_grad(self):
         # Todo: report clipped and unclipped
         # Todo: allow clip=None but still report grad_norm
         grad_clips = self.gradient_clipping
-        return torch.nn.utils.clip_grad_norm_(
-            self.parameters, grad_clips
-        )
+        if isinstance(self.parameters[0], dict) and 'params' in self.parameters[0]:
+            params = itertools.chain(*[param_group['params'] for param_group in self.parameters])
```
boeddeker, line 42: Is that a bug in PyTorch, that the optimizers support more inputs than the `clip_grad_norm_`?
Could you add a comment here, which special case is solved?

padertorch/utils.py
```diff
@@ -142,6 +143,20 @@ def to_numpy(array, detach: bool = False, copy: bool = False) -> np.ndarray:
         if detach:
             array = array.detach()
 
+    if isinstance(array, torch.Tensor) and array.is_sparse:
+        if force_dense:
+            array = array.to_dense()
+        else:
+            raise NotImplementedError
```
boeddeker, line 150: Why do you raise a NotImplementedError? Will Sparce Tensors not be converted to scipy? Or are the ambiguities?

PR #160, opened 2024-06-18
prepare for pypi

boeddeker:


boeddeker:
For documentation: git dependencies aren't allowed for pypi

thequilo (approved):

